PR #151, opened 2019-03-11
updated protocols in line with content document

joefitter:


client/application.js
```diff
@@ -940,23 +836,19 @@ export default {
...
                 type: 'radio',
                 options: [
                   {
                     label: 'Yes',
                     value: true,
                     reveal: {
-                      name: 'continued-use-sourced',
-                      label: 'Where will these animals be sourced from?',
-                      hint: 'This could be another protocol in this licence, another project licence, or somewhere else.',
+                      name: 'continued-use-start',
```
lennym: Can we keep this field name consistent?

joefitter: ooh good spot - I thought I caught them all

PR #152, opened 2019-03-12
check value changed before rendering ALL THE FIELDS

joefitter:


client/pages/sections/protocols/index.js
```diff
@@ -9,7 +9,7 @@ import Fieldset from '../../../components/fieldset';
 import Repeater from '../../../components/repeater';
 import Controls from '../../../components/controls';
 
-class Form extends Component {
+class Form extends PureComponent {
```
lennym: What is this?

joefitter: PureComponent calls shouldComponentUpdate and performs a shallow comparison, only rendering if something changed, Component doesn't run a comparison and renders by default

PR #153, opened 2019-03-12
performance improvements - dont connect field component to project - …

joefitter:
…just take the things we need

### DIFF
--- a/client/components/field.js
+++ b/client/components/field.js
@@ -1,6 +1,5 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
-import { Link } from 'react-router-dom';
 
 import castArray from 'lodash/castArray';
 import every from 'lodash/every';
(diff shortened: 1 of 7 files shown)


client/components/field.js
```diff
@@ -31,7 +27,7 @@ class Field extends Component {
     return this.props.onSave && this.props.onSave(value);
   }
 
-  mapOptions(options = []) {
+  mapOptions = (options = []) => {
```
lennym: Any reason this one function has changed syntax from the rest of the file?

lennym:
I like this. One small question.

client/components/field.js
joefitter: I believe it needs to be bound to `this`, I didn't actually check. We were binding to `this` in a constructor previously, so I deleted that

joefitter: disregard previous comment - it needs to be bound to `this`

lennym: Surely it's already bound to `this` on the basis that it's a class method?

Or is it because we do `someprop={this.mapOptions}` somewhere else (and thus unbinding)?

joefitter: I'll check :)

lennym: It makes sense (hence my merging)

My preference is generally to arrow-function-ify at prop declare time i.e. `someProp={() => this.mapOptions()}`

joefitter: If you do that you are redefining the function every time the child is rendered - `() => {} !== () => {}` so it will appear the prop has changed and trigger a re-render. We should use handle functions when passing a function as a prop:

```js
handleDoSomething = val => {
  this.setState({ val });
}

<Component
  doSomething={handleDoSomething}
/>
```

PR #155, opened 2019-03-13
Feature/protcol review

joefitter:


client/pages/sections/protocols/steps.js
```diff
@@ -1,5 +1,6 @@
 import React, { Component, Fragment } from 'react';
 import { connect } from 'react-redux';
+import { withRouter } from 'react-router';
```
lennym: If we're importing from `react-router` then it should be declared as a dependency in package.json.

client/pages/sections/protocols/steps.js
```diff
@@ -37,16 +38,25 @@ class Step extends Component {
     this.props.moveDown();
   }
 
+  componentDidMount() {
+    if (this.props.parts && this.props.parts.length === 5) {
+      const activeStep = parseInt(this.props.parts[3], 10);
```
lennym: There's a lot of magic numbers here. What's the significance of 5 & 3?

client/pages/sections/protocols/steps.js
joefitter: yes, yes we should. Didn't somebody write an eslint plugin exactly for this purpose  🤔

client/pages/sections/protocols/steps.js
joefitter: I was hoping you wouldn't notice that...

lennym:
I'm not going to pretend that I can completely follow what's going on here, but there's no huge alarm bells.

The only thing that concerns me is that a lot of the time that `parts` is being worked with there's seemingly arbitrary indexes being used to work with it. It might be good to either make the code more explanatory as to what's going on, or add some comments.

client/pages/sections/protocols/protocols.js
```diff
@@ -0,0 +1,128 @@
...
+  toggleActive = () => {
+    this.setState({ active: !this.state.active });
+  }
+
+  getParts = () => {
+    if (!this.props.editable || !this.props.location.hash) {
+      return null;
+    }
+    const parts = this.props.location.hash.split('.');
+    if (!parts.length) {
```
lennym: I don't _think_ this is possible here. I think worst case scenario you end up with `['']`.

client/pages/sections/protocols/protocols.js
```diff
@@ -0,0 +1,128 @@
...
+      return null;
+    }
+    return parts;
+  }
+
+  isActive = parts => {
```
lennym: The only time we call this function we have called `this.getParts` immediately before to find the argument. 

I'd probably move that call to inside this function and do away with the argument here for marginally improved efficiency.

lennym: See edit.

client/pages/sections/protocols/protocols.js
```diff
@@ -0,0 +1,128 @@
...
+    this.props.save({ protocols });
+  }
+
+  edit = e => {
+    e.preventDefault();
+    this.props.retreat();
+  }
+
+  render() {
+    const { project, editable } = this.props;
```
lennym: It doesn't look like we need anything other than `project.protocols` here to actually render, so is there a performance benefit to only returning the protocols from `mapStateToProps`?

client/pages/sections/protocols/steps.js
lennym: It doesn't understand `import` because someone is too lazy to update it.

client/pages/sections/protocols/steps.js
lennym: 👀

client/pages/sections/protocols/protocols.js
lennym: I've tried to trace through the downstream components that get rendered with `{...props}` and none of them seem to obviously be using the project.

joefitter: right you are

client/pages/sections/protocols/protocols.js
joefitter: good spot

PR #160, opened 2019-03-20
Content changes

pcuk1979:
All changes to questions per protocol, plus a few extra changes elsewhere that I could easily deal with myself.

### DIFF
--- a/client/application.js
+++ b/client/application.js
@@ -952,7 +952,7 @@ export default {
               },
               {
                 name: 'code',
-                label: 'Select the anaesthetic codes that apply to this step?',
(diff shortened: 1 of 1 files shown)


pcuk1979:
May be a few errant { and } in here. Apols.

client/application.js
```diff
@@ -952,7 +952,7 @@ export default {
               },
               {
                 name: 'code',
-                label: 'Select the anaesthetic code you intend to use.',
+                label: 'Select the anaesthetic codes that apply to this step?',
```
lennym: The question mark doesn't look right here. It's not a question?

pcuk1979: Replace it with a .

PR #161, opened 2019-03-20
Bugfix/render reveals

joefitter:


client/components/fieldset.js
```diff
@@ -11,6 +11,8 @@ const Fieldset = ({ fields, onFieldChange, values, prefix = '' }) => (
           key={ field.name }
           name={ `${prefix}${field.name}` }
           value={ values && values[field.name] }
+          values={ values }
+          prefix={ prefix }
```
lennym: What are these new props for?

PR #163, opened 2019-03-22
Legacy PPLS

joefitter:
Added aspel schema, altered application reducer to control schemaVersion and readonly props. These are passed in by the implementation.

Added aspel specific legacy animals section for protocols


client/schema/index.js
```diff
@@ -0,0 +1,7 @@
+import aspel from './aspel';
+import marvell from './marvell';
```
lennym: Can we rename these to `v0` and `v1`?

PR #197, opened 2019-04-09
Feature/pdf

joefitter:


client/pdf/protocols/index.js
```diff
@@ -0,0 +1,85 @@
...
+      <ReviewTextEditor value={1} />
+
+      <p className="end">{`( End of protocol ${number} )`}</p>
+    </div>
+  </div>
+)
+
+const PDF = ({ protocols = [], schemaVersion }) => {
+  if (schemaVersion !== 0) {
+    console.error('PDF generation is currently only supported for legacy PPLs');
```
lennym: We might want to throw this error rather than just logging it.

joefitter: yeah that's a good idea, I was going to display it in the red "toaster" then I remembered that isn't a thing

PR #198, opened 2019-04-12
added pdf to publish list

joefitter:


lennym:
I'd normally go for an `.npmignore` file rather than a whitelist unless there's a good reason for exactly this reason. It's really easy to footgun yourself by forgetting to add files.

joefitter:
yeah, as demonstrated

PR #205, opened 2019-04-25
Per question commenting

joefitter:
* Added comments reducer and actions
* Added comments and add comment components
* Added new comments badge
* Added comments summary to application summary at a section level
* Added to protocols at a protocol level


client/components/application-summary.js
```diff
@@ -73,6 +99,29 @@ class ApplicationSummary extends React.Component {
...
 
+  getComments = subsection => {
+    let newComments = 0;
+
+    const repeaters = {
+      protocols: /^protocol\./,
+      establishments: /^establishments\./,
+      poles: /^pole\./,
+      'action-plan': /objectives\./
+    }
```
lennym: Is it possible to extract this from the schema? I'm mildly concerned it might differ across schema versions.

PR #208, opened 2019-04-29
Update index.js

pcuk1979:
PR with potential para break in it.

### DIFF
--- a/client/schema/v1/index.js
+++ b/client/schema/v1/index.js
@@ -899,9 +899,7 @@ each other.`,
                 name: 'title',
                 type: 'texteditor',
                 label: 'Describe the procedures that will be carried out during this step.',
-                hint: `Explain where one or more steps are repeated in one experiment, list any alternative techniques within a step (e.g. dosing routes), and include all procedures performed under terminal anaesthesia.
-
-                When describing the technical aspects of a step, be broad enough to be flexible when the variation does not impact on animal welfare (e.g. use "antibiotic" instead of "penicillin").  Finally, avoid specifying  volumes and frequencies when they do not impact on animal welfare.`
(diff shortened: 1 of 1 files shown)


client/schema/v1/index.js
```diff
@@ -892,14 +892,16 @@ each other.`,
...
             footer: 'Once you’ve created a list of steps, you need to add information about adverse effects, controls and limitations, and humane endpoints to each one.​',
             fields: [
               {
                 name: 'title',
                 type: 'texteditor',
                 label: 'Describe the procedures that will be carried out during this step.',
-                hint: 'To ensure that an adequate harm benefit assessment can be carried out, please provide a clear and explicit explanation of each step.'
+                hint: `Explain where one or more steps are repeated in one experiment, list any alternative techniques within a step (e.g. dosing routes), and include all procedures performed under terminal anaesthesia.
+
+                When describing the technical aspects of a step, be broad enough to be flexible when the variation does not impact on animal welfare (e.g. use "antibiotic" instead of "penicillin").  Finally, avoid specifying  volumes and frequencies when they do not impact on animal welfare.`
```
lennym: So this has not worked, but it's not worked in a very curious way.

Will need to take a closer look.

client/schema/v1/index.js
```diff
@@ -825,8 +825,8 @@ each other.`,
               },
               {
                 name: 'maximum-times-used',
-                label: 'What is the maximum number of times that this protocol will be used on this type of animal?',
-                hint: 'Count the number of first time uses of a naive animal, then add this to the number of instances of re-use (if relevant).',
+                label: 'What is the maximum number of uses of this protocol on this type of animal??',
```
lennym: Double question mark??

PR #209, opened 2019-04-29
Mark changes on PPL version in pink

lvnhmd:


assets/scss/projects.scss
```diff
@@ -269,6 +269,10 @@ $light-green: #26A197;
     &.comments {
       background: $light-green;
     }
+
+    &.changed {
+      background: #ff1493;
```
joefitter: could we add a variable please?

client/components/application-summary.js
```diff
@@ -95,6 +97,24 @@ class ApplicationSummary extends React.Component {
     }
   }
 
+  changed = (key) => {
+    const fields = this.props.fieldsBySection[key] || [];
+    if(this.props.changed.includes(key) || this.props.changed.some(k=> fields.includes(k) )) {
+      return CHANGED;
+    }
+  }
```
joefitter: @lvnhmd I think you'll also need to include looping sections - see the use of `repeats` property in the comments code

assets/scss/projects.scss
lvnhmd: done

client/components/application-summary.js
lvnhmd: done

joefitter:
I have boueuouf (:cow2:) with adding 2 no-op reducers for the same thing, more so with the naming of them. Could we have a single reducer, called `amends`. Also this only shows amends on first page load, we'll probably want to show live changes as a user is amending fields

client/components/application-summary.js
```diff
@@ -95,6 +98,27 @@ class ApplicationSummary extends React.Component {
     }
   }
 
+  changed = (key, subsection) => {
+    const fields = this.props.fieldsBySection[key] || [];
+    if(this.props.latest.includes(key) || this.props.latest.some(k=> fields.includes(k)) || this.props.latest.includes(subsection.repeats)) {
```
joefitter: `if (`
`k =>`
spaces please - not really your fault - this should be picked up by lint - I guess it's ignoring JSX

joefitter: actually that's a lie - this isn't a jsx file - our linting is obvs borked

PR #210, opened 2019-04-29
bugfixes for legacy comment, filter own comments from new labels, sho…

joefitter:
…w new labels in protocols substeps

### DIFF
--- a/client/components/comments/index.js
+++ b/client/components/comments/index.js
@@ -63,10 +63,10 @@ class Comments extends Component {
       >
         <Fragment>
           {
-            showPrevious && previous.map((comment, index) => <Comment index={index} key={index} { ...comment } user={this.props.user} />)
+            (showPrevious || !active.length) && previous.map((comment, index) => <Comment index={index} key={index} { ...comment } user={this.props.user} />)
           }
           {
-            !!previous.length && (
+            !!active.length && !!previous.length && (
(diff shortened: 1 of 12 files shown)


client/pages/sections/establishments/index.js
```diff
@@ -1,4 +1,5 @@
 import React, { Component } from 'react';
+import { v4 } from 'uuid';
```
lennym: You should import from `uuid/v4` to avoid including all the code for the other versions.

joefitter: good shout

client/components/comments/index.js
```diff
@@ -63,10 +63,10 @@ class Comments extends Component {
       >
         <Fragment>
           {
-            showPrevious && previous.map((comment, index) => <Comment index={index} key={index} { ...comment } user={this.props.user} />)
+            (showPrevious || !active.length) && previous.map((comment, index) => <Comment index={index} key={index} { ...comment } user={this.props.user} />)
           }
           {
-            !!previous.length && (
+            !!active.length && !!previous.length && (
               <p><Button className="link" onClick={this.toggleShowPrevious}>{ showPrevious ? 'Hide' : 'Show' }</Button> {previous.length} previous comments</p>
```
lennym: Out of scope I know, but is there a reason that _only_ the hide/show part is clickable? That seems super weird to me, and contrary to all received wisdom.

PR #215, opened 2019-05-01
animal quantities were missing in docx - added custom renderer

joefitter:


client/components/download-link/renderers/docx.js
```diff
@@ -322,6 +359,10 @@ const renderField = (doc, field, values) => {
     return renderSpeciesSelector(doc, values, value);
   }
 
+  if (field.type === 'animal-quantities') {
+    return renderAnimalQuantities(doc, values);
+  }
```
lennym: Does this need to move below the `renderNull` block below?

lennym: No.

joefitter: /shrug.gif

PR #217, opened 2019-05-01
added delete link to own comments, added ajax action

joefitter:


client/actions/comments.js
```diff
@@ -41,6 +49,28 @@ const postData = (comment, getState, dispatch) => {
...
+    .then(response => {
+      return response.json()
+        .then(json => {
+          if (response.status > 399) {
+            const err = new Error(json.message || `Failed delete comment with status code: ${response.status}`);
+            err.status = response.status;
+            Object.assign(err, json);
+            throw err;
+          }
+        })
```
lennym: Is there a case for abstracting some of this boilerplate?

joefitter: yes, absolutely. Is it the case that I was just being extremely lazy? Also yes.

PR #220, opened 2019-05-03
we were updating the state async in repeater, which is all good and w…

joefitter:
…ell until something else updates it sync, and the whole world ends.

### DIFF
--- a/client/components/repeater.js
+++ b/client/components/repeater.js
@@ -53,10 +53,7 @@ class Repeater extends Component {
   }
 
   update(items) {
-    return new Promise(resolve => {
-      this.setState({ items }, resolve)
-    })
-      .then(this.save);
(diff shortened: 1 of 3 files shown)


client/components/repeater.js
```diff
@@ -53,10 +53,7 @@ class Repeater extends Component {
   }
 
   update(items) {
-    return new Promise(resolve => {
-      this.setState({ items }, resolve)
-    })
-      .then(this.save);
+    this.setState({ items }, this.save)
```
lennym: How is this different? The code seems the same to me.

joefitter: it was wrapped in a promise

joefitter: therefore deferred

PR #221, opened 2019-05-03
defer addChange

joefitter:


lennym:
Closed in favour of #222

PR #224, opened 2019-05-08
Add changed badge to questions

lvnhmd:


client/components/review.js
```diff
@@ -125,10 +126,34 @@ class Review extends React.Component {
...
+        return null;
+    }
+  }
+
   render() {
     return (
       <div className="review">
         <h3>{this.props.label}</h3>
+        {
+          this.changedBadge(this.changed(`${this.props.prefix || ''}${this.props.name}`))
```
lennym: Since all the arguments here are properties of `this` I'd just look up the key directly in `this.changedBadge()` rather than passing them around as arguments.

lennym: That is, add 

```
const changed = this.changed(`${this.props.prefix || ''}${this.props.name}`)
``` 

to the top of `this.changedBadge()` and get rid of the argument.

client/components/review.js
```diff
@@ -149,6 +174,6 @@ class Review extends React.Component {
   }
 }
 
-const mapStateToProps = ({ project, settings, application: { readonly } }) => ({ project, settings, readonly });
+const mapStateToProps = ({ project, settings, application: { readonly }, changes : {latest, granted} }) => ({ project, settings, readonly, latest, granted });
```
lennym: It might be better to traverse the changes in this function using the props provided and _only_ pass the changed state for this one question rather than the entire list.

The reasoning being that if the entire list is passed then this field will re-render whenever any change is made, not just the one field that we care about.

Having said that, I don't think it matters _too much_ here because it's static data, but worth considering for future components.

client/components/review.js
```diff
@@ -10,6 +10,7 @@ import flatten from 'lodash/flatten';
 import values from 'lodash/values';
 import isUndefined from 'lodash/isUndefined';
 import isNull from 'lodash/isNull';
+import { LATEST, GRANTED } from '../constants/change';
```
lennym: Are these constants actually used anywhere else? It seems like they're just references within this one file, so I probably wouldn't abstract them away like this.

client/components/review.js
lennym:
```js
const mapStateToProps = ({ project, settings, application: { readonly }, changes : {latest, granted} }, { name, prefix }) => {
  const key = `${this.props.prefix || ''}${this.props.name}`;
  const changedFromGranted = granted.includes();
  const changedFromLatest = latest.includes();
  return { project, settings, readonly, changedFromLatest, changedFromGranted };
}
```

client/components/review.js
lvnhmd: yes, in application-summary

client/components/review.js
lvnhmd: thank you

client/components/review.js
```diff
@@ -125,10 +126,33 @@ class Review extends React.Component {
...
+
+  changedBadge = () => {
+    switch (this.changed()) {
+      case LATEST:
+        return <span className="badge changed">changed</span>;
+      case GRANTED:
+        return <span className="badge">amended</span>;
+      default:
+        return null;
+    }
```
lennym: There's a lot of misdirection in this code now, calling out to another function and importing constants, to just do a simple condition on a boolean prop. 

I think we can reduce this function to:

```
changedBadge = () => {
  if (this.props.changedFromLatest) {
    return <span className="badge changed">changed</span>;
  }
  if (this.props.changedFromGranted) {
    return <span className="badge">amended</span>;  
  }
  return null;
}
```

And then get rid of `this.changed()` and the constants.

client/components/review.js
```diff
@@ -125,10 +126,33 @@ class Review extends React.Component {
...
+  }
+
   render() {
     return (
       <div className="review">
         <h3>{this.props.label}</h3>
+        {
+          this.changedBadge()
+        }
         {this.replay()}
```
lennym: Can we just make the formatting consistent here? Either have the line breaks inside the `{  }` or don't. Either is fine, but formatting the "same" code two different ways right next to each other makes me twitch.

PR #233, opened 2019-05-17
Added other legal text section

joefitter:
* updated conditions and authoristions in line with spreadsheet
* added other legal text section
* abstracted editable field component

### DIFF
--- a/client/actions/projects.js
+++ b/client/actions/projects.js
@@ -106,21 +106,54 @@ export function updateAndSave(data) {
   };
 }
 
+const doConditionsUpdate = (data, dispatch, state) => {
+  const { application: { basename } } = state;
+  const params = {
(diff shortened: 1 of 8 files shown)


client/actions/projects.js
```diff
@@ -106,21 +106,54 @@ export function updateAndSave(data) {
...
+  const { application: { basename } } = state;
+  const params = {
+    method: 'PUT',
+    url: `${basename.replace(/\/edit?/, '')}/conditions`,
+    data
+  };
+  return Promise.resolve()
+    .then(() => sendMessage(params))
+    .catch(err => {
+      console.error(err);
```
lennym: Do we care about this? I suspect it might actually be useful so I'm happy to leave it.

PR #234, opened 2019-05-17
Spike diffing

lennym:


lennym:
I'mma close this so it disappears from the build monitor, but leave the code in place.

PR #236, opened 2019-05-21
Do not show 'revert to default' for custom conditions and authorisations

lvnhmd:


joefitter:
@lvnhmd this component is also used for the other-legal-text section - you'll need to add it there too

joefitter:
pretty sure this still isn't going to work @lvnhmd - you introduced a new prop `showRevert` in the component, it is falsy by default, therefore we wont show the revert component in edited other-legal-text. You'll either need to add the showRevert prop there too, or perhaps default it to `true` in the component.

lvnhmd:
> pretty sure this still isn't going to work @lvnhmd - you introduced a new prop `showRevert` in the component, it is falsy by default, therefore we wont show the revert component in edited other-legal-text. You'll either need to add the showRevert prop there too, or perhaps default it to `true` in the component.

It does not seem that we save a property 'custom' when adding other legal text, because we  don't have the ability to add a custom other legal text. I am not sure what the value of `showRevert` should be.

joefitter:
just default it to true

joefitter:
if you don't then this PR breaks current functionality.

lvnhmd:
> just default it to true

Done.

PR #249, opened 2019-06-04
Show continuation info for legacy projects

wheelsandcogs:


client/components/review.js
```diff
@@ -37,6 +38,11 @@ class Review extends React.Component {
         </dl>
       )
     }
+
+    if (value && this.props.type === 'date') {
+      return formatDate(value, 'DD/MM/YYYY');
```
lennym: Can we use standard date formats from constants pls? It might be worth coming up with a date-format component somewhere, but that is likely out of scope.

PR #255, opened 2019-06-07
Make the continuation a reveal

wheelsandcogs:


client/schema/v0/index.js
```diff
@@ -21,15 +21,36 @@ export default {
...
-            name: 'continuation-licence-number',
-            label: 'Project continuation - previous licence number',
-            type: 'text'
-          },
-          {
-            name: 'continuation-expiry-date',
-            label: 'Project continuation - previous licence expiry date',
-            type: 'date'
+            name: 'continuation',
+            label: 'Is this a continuation of a previous project licence that has expired?',
```
lennym: Has this content been provided by a Phil that isn't you?

wheelsandcogs: Sort of - he provided "This is a continuation of a project licence that has expired." but it needs to be a question.

wheelsandcogs: Looks like I've added a "previous" from out of nowhere - I'll remove it.

lennym: In which case, we might want to be careful about "that has expired", there's a small but real chance that it's a continuation of a project that has not expired, but will expire soon, which makes the question potentially quite ambiguous.

In particular in cases where an applicant is applying for a new continuation. The previous project will _necessarily_ not have expired at that point (and in before "but legacy only" - we're migrating draft applications so this could still apply).

Paging @pcuk1979 for an opinion.

pcuk1979: If it needs to be a question, just go with

Is this a continuation of a previous project licence?

wheelsandcogs: ta!

PR #259, opened 2019-06-12
Add superscript and subscript to RTE

lvnhmd:


client/components/editor/text-editor.js
```diff
@@ -33,6 +35,8 @@ const isBoldHotkey = isKeyHotkey('mod+b');
 const isItalicHotkey = isKeyHotkey('mod+i');
 const isUnderlinedHotkey = isKeyHotkey('mod+u');
 const isCodeHotkey = isKeyHotkey('mod+`');
+const isSuperscriptHotkey = isKeyHotkey('mod+=');
+const isSubscriptHotkey = isKeyHotkey('mod+shift++');
```
wheelsandcogs: just out of interest, are these default hotkeys or hotkeys we've chosen?

lvnhmd: they seem to be the industry accepted hotkeys i.e MSWord ones

wheelsandcogs: 👍

PR #261, opened 2019-06-18
Added slate-edit-table

joefitter:
* added @joefitter/slate-edit-table forked from Gitbook.IO and patched to work with state@0.47
* refactored editor into plugins


package.json
```diff
@@ -67,8 +68,11 @@
     "redux-logger": "^3.0.6",
     "redux-thunk": "^2.3.0",
     "regenerator-runtime": "^0.13.2",
-    "slate": "^0.44.9",
-    "slate-react": "^0.21.16",
+    "slate": "^0.47.4",
+    "slate-edit-list": "^0.12.1",
+    "slate-edit-table": "^0.18.0",
```
lennym: Do we need `slate-edit-table` *and* `@joefitter/slate-edit-table`?

lennym:
It is my anticipation that this will comprehensively knacker Word downloads. I am also ok with that right now.

package.json
joefitter: nope, we don't need the list one either (yet) - good spot.

PR #267, opened 2019-06-26
Feature - other establishments

joefitter:
* Removed poles and establishments views and reviews
* Added generic repeating step and review
* updated config to use new repeater


lennym:
Does this contain any of the code we discussed to normalise an establishment list containing string values into an establishment list of objects?

That is, is it backwards compatible with data created before this change, and without the migrator change in https://github.com/UKHomeOffice/asl-migration/pull/96

joefitter:
the migrator change is needed as before it was just wrong, we were mapping the values to a different field - `other-establishments-list` was a checkbox list containing just names and `establishments` is the array of objects

joefitter:
The only b/c case we need to think about is when somebody had previously selected establishments from the checkbox list but not filled out any details for them. This is because they were previously mapped and the names added on the next step.  This change assumes if there are establishments, then the data is ok, if not it will create them from the list of names

PR #279, opened 2019-07-15
there is no primary establishment in drafting tool land

joefitter:


client/components/download-link/renderers/docx.js
```diff
@@ -288,7 +288,7 @@ export default application => {
 
     if (field.type === 'location-selector') {
       const options = [
-        application.establishment.name,
+        ...(application.establishment && [application.establishment.name] || []),
```
lennym: That's not a ternary, but I think you meant it to be.

client/components/location-selector.js
```diff
@@ -16,7 +16,7 @@ const LocationSelector = ({
       type="checkbox"
       className="smaller"
       options={[
-        establishment.name,
+        ...(establishment && [establishment.name] || []),
```
lennym: Same.

client/pages/readonly.js
```diff
@@ -11,14 +11,14 @@ function Readonly({ establishment, isGranted, project, options, schemaVersion, s
   return (
     <Fragment>
       {
-        isGranted && <h3 className="est-title">{establishment.name}</h3>
+        isGranted && <h3 className="est-title">{establishment && establishment.name}</h3>
```
lennym: `establishment &&` probably wants pulling outside the `<h3>` so it only renders the tag if it exists.

lennym:
```suggestion
        establishment && isGranted && <h3 className="est-title">{establishment.name}</h3>
```

PR #280, opened 2019-07-15
Bugfix - text to RTE

joefitter:
If plain text is passed in to a RTE field a JSON parse error is thrown. If value, try to parse json, if not then set as initial text


client/components/review.js
```diff
@@ -140,7 +140,7 @@ class Review extends React.Component {
       </dl>
     }
     if (this.props.type === 'texteditor' && this.props.value) {
-      if (this.props.value !== JSON.stringify(Value.fromJSON(initialValue).toJSON())) {
+      if (this.props.value !== JSON.stringify(Value.fromJSON(initialValue('')).toJSON())) {
```
wheelsandcogs: can we use `notNull` here as well to make the code clearer? ta.

lennym: I see it's out of scope, but what's the deal with `.fromJSON(.toJSON())`? Feels redundant.

Normalisation of json structure across Slate versions perhaps?

joefitter: yes that, extra properties are added to the structure when you pass it to Value.fromJSON

PR #286, opened 2019-07-15
remove filter from review fields - this was preventing some fields re…

joefitter:
…ndering

### DIFF
--- a/client/components/review-fields.js
+++ b/client/components/review-fields.js
@@ -17,7 +17,7 @@ const fieldIncluded = (field, values) => {
   return every(Object.keys(field.conditional), key => field.conditional[key] === values[key])
 }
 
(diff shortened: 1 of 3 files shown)


lennym:
It's probably worth defaulting the values to `{}` in a `ReviewFields` and `Fieldset` component as well

client/components/review-fields.js
```diff
@@ -20,7 +20,7 @@ const fieldIncluded = (field, values) => {
 const ReviewFields = ({ fields, values, onEdit, editLink, project, prefix = '', noComments, altLabels, readonly, step }) => (
```
lennym:
```suggestion
const ReviewFields = ({ fields, values = {}, onEdit, editLink, project, prefix = '', noComments, altLabels, readonly, step }) => (
```

lennym: Just in case you missed some.

PR #295, opened 2019-07-18
Legacy PDFs

joefitter:
* updated to show all sections
* show standard conditions on both licence types


client/components/review.js
```diff
@@ -24,10 +24,12 @@ class Review extends React.Component {
 
   render() {
     const { label } = this.props.altLabels ? this.props.alt : this.props;
-    const { hint } = this.props;
+    const { hint, isGranted, hideGrantedLabel } = this.props;
     return (
       <div className="review">
-        <h3>{label}</h3>
+        {
+          (!isGranted || !hideGrantedLabel) && <h3>{label}</h3>
```
wheelsandcogs: This would probably be more readable if the prop was `showGrantedLabel` to avoid the double negative :)

PR #302, opened 2019-07-24
Implement granular diffs in comparison tooling

lennym:
Compares the string values of the before and after fields and highlights removals and additions on each.


joefitter:
🎉

PR #304, opened 2019-07-25
Nested lists

joefitter:
* added slate-edit-list fork
* configured plugin

### DIFF
--- a/assets/scss/_editor.scss
+++ b/assets/scss/_editor.scss
@@ -59,6 +59,9 @@
     &.active {
       color: $blue;
     }
+    &[disabled] {
+      color: $mid-grey;
+    }
   }
   img {
     display: block;
@@ -183,5 +186,18 @@
 .review {
   > [data-slate-editor] {
     pointer-events: none;
+  }
+}
+
+.current-item {
+  position: relative;
+
+  &:after {
+    position: absolute;
+    content: '';
+    border: 2px solid #d9edff;
+    left: -25px;
+    height: 100%;
+    top: 0;
   }
 }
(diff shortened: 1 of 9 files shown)


client/components/editor/index.js
```diff
@@ -67,16 +67,12 @@ class TextEditor extends Component {
   };
 
   onBlur = (self, editor, next) => {
-    next();
+    next()
```
wheelsandcogs: ;

wheelsandcogs:
```suggestion
    next();
```

joefitter: there are about a billion of these in projects, as our linting doesn't care, we should make it care, but let's do them all together?

assets/scss/_editor.scss
```diff
@@ -185,3 +188,16 @@
...
+  position: relative;
+
+  &:after {
+    position: absolute;
+    content: '';
+    border: 2px solid #d9edff;
+    left: -25px;
+    height: 100%;
+    top: 0;
+  }
```
lennym: What's occurring here?

joefitter: its something I took from the example in the plugin - we may or may not want it, but it seems kind of useful when editing a list/nested list - it highlights the currently scoped list item on which the change will be performed

joefitter: see here https://gitbookio.github.io/slate-edit-list/

joefitter: happy to remove it if you think it's not needed

lennym: No, I was just curious what it was for. Looks nice to me.

PR #305, opened 2019-07-26
Feature/ppl export

joefitter:


client/components/download-header.js
```diff
@@ -1,14 +1,64 @@
...
+  const container = useRef(null);
+  const download = useRef(null);
+  basename = basename.replace(/\/edit/, '');
+
+  // project title could span multiple lines, adjust download position accordingly
+  useEffect(() => {
+    // subtract padding, and border
+    const height = container.current.offsetHeight - 30 - 4;
+    download.current.style.height = `${height}px`;
+    download.current.style.lineHeight = `${height}px`;
```
lennym: Is this setting inline styles? Does that actually work with CSP policies as they are?

lennym: Spoiler: it shouldn't.

joefitter: it doesn't, updating the dom element in this way is fine with CSP, it's only if you set inline styles on the element using the style attribute

joefitter: also this is client side, so CSP has already done it's shizzle by the time this runs

lennym: I'm not sure that last bit is true - CSP should prevent new elements which violate the policy being added, but if it works then I'm happy.

joefitter: ORLY? I did not know that. However this does work

PR #319, opened 2019-08-02
Docx tables

joefitter:
* create matrix representaion of table including colspans and rowspans
* create table with max dimensions
* add content to cells
* pass over matrix again merging rowspans - this needs to happen before horizontal merges so these need to be merged in a separate pass


client/components/download-link/renderers/docx-renderer.js
```diff
@@ -119,6 +121,104 @@ export default (application, sections, values) => {
     return doc;
   };
 
+  const renderTable = (doc, node) => {
+    const tableToMatrix = table => {
```
lennym: I'd be tempted to pull this function up a level if it doesn't need the `doc` or `node` variables in scope (and I don't think it does). Purely to make the code more readable.

joefitter: right you are

PR #326, opened 2019-08-05
Added field schema for legacy experience

wheelsandcogs:


client/schema/v0/index.js
```diff
@@ -325,6 +353,7 @@ export default {
           },
           animals: {
             title: 'Type of animals',
+            repeats: 'species',
```
lennym: What's going on here?

This seems pretty wildly out of scope.

wheelsandcogs: it was already in @joefitter's commit, I assume it's a fix for something else?

PR #332, opened 2019-08-07
if displaying a readonly texteditor, empty values should display No a…

joefitter:
…nswer provided

### DIFF
--- a/client/components/editor/index.js
+++ b/client/components/editor/index.js
@@ -27,6 +27,14 @@ const plugins = [
   listPlugin
 ];
 
+const serialiseValue = value => {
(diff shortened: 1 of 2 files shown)


client/components/editor/index.js
```diff
@@ -27,6 +27,14 @@ const plugins = [
   listPlugin
 ];
 
+const serialiseValue = value => {
+  const jsonVal = JSON.stringify(value.toJSON());
+  if (jsonVal === JSON.stringify(Value.fromJSON(initialValue('')))) {
```
lennym: I _think_ this is ok as:
```suggestion
  if (jsonVal === initialValue('')) {
```

PR #333, opened 2019-08-07
Add legacy establishment component

lvnhmd:


client/pages/sections/legacy-establishments.js
```diff
@@ -0,0 +1,36 @@
+
+import React, {Fragment} from 'react';
+import { connect } from 'react-redux';
+import RepeaterReview from './repeater/review';
+
+const LegacyEstablishments = ({ title, steps, values, readonly, singular, enable,
```
lennym: Either line-breaks or no line-breaks, but not this half-way house.

client/pages/sections/legacy-establishments.js
```diff
@@ -0,0 +1,36 @@
...
+          <dt>Primary establishment: </dt>
+          <dd>{ name }</dd>
+          <dt>Establishment licence number: </dt>
+          <dd>{ licenceNumber }</dd>
+          <dt>Establishment licence holder: </dt>
+          <dd>{firstName} {lastName}</dd>
+          <dt>Address: </dt>
+          <dd>{ address }</dd>
+        </dl>
+      <RepeaterReview title={title} steps={steps} values={values} readonly={readonly} singular={singular} enable={enable} />
```
joefitter: `<RepeaterReview {...props} />`

client/pages/sections/legacy-establishments.js
joefitter: `({ establishment, ...props })`

client/pages/sections/legacy-establishments.js
```diff
@@ -0,0 +1,36 @@
...
+    </Fragment>
+  )
+}
+
+const mapStateToProps = ({ project, application: { readonly, establishment } }) => ({ project, readonly, establishment });
```
joefitter: we only need to map the establishment here, not props or readonly as this component doesn't care about them

client/pages/sections/legacy-establishments.js
```diff
@@ -0,0 +1,36 @@
...
+const LegacyEstablishments = ({ title, steps, values, readonly, singular, enable,
+          establishment: {
+            name,
+            licenceNumber,
+            address,
+            licenceHolder: {
+              firstName,
+              lastName
+            }
+          }
```
joefitter: I assume this indentation will be fixed :)

client/pages/sections/legacy-establishments.js
```diff
@@ -0,0 +1,36 @@
...
+import React, {Fragment} from 'react';
+import { connect } from 'react-redux';
+import RepeaterReview from './repeater/review';
+
+const LegacyEstablishments = ({ title, steps, values, readonly, singular, enable,
+          establishment: {
+            name,
+            licenceNumber,
+            address,
+            licenceHolder: {
```
lennym: You will need to handle the case where this is not set.

client/pages/sections/legacy-establishments.js
```diff
@@ -0,0 +1,29 @@
...
+  return (
+    <Fragment>
+      <h2>Primary establishment</h2>
+      <dl className="inline">
+      <dt>Establishment name: </dt>
+      <dd>{ name }</dd>
+      <dt>Licence number: </dt>
+      <dd>{ licenceNumber }</dd>
+      </dl>
+      <br/>
```
lennym: This feels like it shouldn't be necessary between block elements. If you want to add whitespace then you should use CSS.

client/pages/sections/legacy-establishments.js
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <Fragment>
+      <div className="granted-section">
+      <h2>Primary establishment</h2>
+      <dl className="inline">
+        <dt>Establishment name: </dt>
+        <dd>{ name }</dd>
+        <dt>Licence number: </dt>
+        <dd>{ licenceNumber }</dd>
+      </dl>
```
lennym: The children of the `<div>` should be indented one more level.

PR #335, opened 2019-08-08
lock docx to rc4 until we work out what is broken

wheelsandcogs:


package.json
```diff
@@ -45,7 +45,7 @@
     "date-fns": "^1.30.1",
     "details-element-polyfill": "^2.3.0",
     "diff": "^4.0.1",
-    "docx": "^5.0.0-rc4",
+    "docx": "5.0.0-rc4",
```
lennym: It's probably worth going back to 4.7.1 and getting off the `rc` train completely.

The major version bump only seemed to come in [here](https://github.com/UKHomeOffice/asl-projects/commit/7871a5bd28c49ab68dd1af67f4be7f79d7b68d3c) when Joe swapped from his forked version (based on 4.7.0) back to core.

If 4.7.1 works ok with the tables, then we should use that.

wheelsandcogs: We've discovered that we'd need to go back to the forked version and we're not sure why Joe switched, so for now just stick to RC4.

PR #337, opened 2019-08-09
Improve syncing

joefitter:
* only allow one sync at a time, keep syncing until everything up to date
* added unbeforeunload if syncing
* disabled next steps if syncing

### DIFF
--- a/client/actions/projects.js
+++ b/client/actions/projects.js
@@ -2,11 +2,10 @@ import omit from 'lodash/omit';
 import debounce from 'lodash/debounce';
 import cloneDeep from 'lodash/cloneDeep';
 import isEqual from 'lodash/isEqual';
-import get from 'lodash/get';
+import pick from 'lodash/pick'
 
 import * as types from './types';
 import database from '../database';
-import { throwError } from './messages';
 import { showMessage } from './messages';
 import sendMessage from './messaging';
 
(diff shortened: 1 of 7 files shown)


client/project-router.js
```diff
@@ -8,11 +9,36 @@ import Project from './pages/project';
...
+  const isSyncing = useSelector(state => state.application.isSyncing);
 
   function toggleStatusShowing() {
     setStatusShowing(!statusShowing);
   }
 
+  useEffect(() => {
+    window.onbeforeunload = () => {
+      if (isSyncing) {
+        return 'Data is currently syncing, if you leave this page your recent changes may not be saved.';
```
lennym:
```suggestion
        return 'Data is currently being saved, if you leave this page your recent changes may not be saved.';
```

lennym:
It's not quite there. With this code as is, then if the conditions are being synced (and so `inSyncing` is true) then a change to `retrospectiveAssessment` is completely disregarded.

I think the check for changed fields need to incorporate all possible fields, and sync _any_ changes if found, rather than just syncing one field at a time.

An alternative approach would be to remove the condition, but move the "is there a sync in progress" condition inside `sync` and add a delay and recursive call in an `else` clause.

client/project-router.js
```diff
@@ -8,11 +9,40 @@ import Project from './pages/project';
...
+      }
+    }
+
+    const nextSteps = document.querySelector('#page-component > p.next-steps > a');
+    const statusMessage = document.querySelector('#page-component > p.next-steps > span.status-message');
+
+    if (isSyncing) {
+      nextSteps.setAttribute('disabled', true);
+      nextSteps.onclick = () => false;
+      statusMessage.innerText = 'Saving...';
```
lennym:
```suggestion
      statusMessage && statusMessage.innerText = 'Saving...';
```

client/project-router.js
```diff
@@ -8,11 +9,40 @@ import Project from './pages/project';
...
+    } else {
+      nextSteps.removeAttribute('disabled');
+      nextSteps.onclick = null;
+      statusMessage.innerText = '';
```
lennym:
```suggestion
      statusMessage && statusMessage.innerText = '';
```

client/project-router.js
```diff
@@ -8,11 +9,40 @@ import Project from './pages/project';
...
+    }
+
+    return () => {
+      window.onbeforeunload = null;
+      nextSteps.removeAttribute('disabled');
+      nextSteps.onclick = null;
+      statusMessage.innerText = '';
```
lennym:
```suggestion
      statusMessage && statusMessage.innerText = '';
```

client/actions/projects.js
```diff
@@ -106,20 +119,47 @@ export function updateAndSave(data) {
...
-      dispatch(throwError('Error updating conditions'));
+      dispatch(doneSyncing());
+      return sync(dispatch, getState, fields, extra);
     });
-}
+};
+
+const debouncedSync = debounce((dispatch, getState, extra) => {
+  return sync(dispatch, getState, ['conditions', 'retrospectiveAssessment'], extra)
+}, 1000, { maxWait: 5000 });
```
wheelsandcogs: Does this mean we do a max of 1 sync per second, with a 5 second timeout?

joefitter: it means it waits for a 1 second pause before trying to sync again, up to a max of 5 seconds. Once the sync is called it starts the recursive dance

PR #339, opened 2019-08-12
Bugfix/as 1559 missing changed label

lvnhmd:
Add 'change' labels on the protocol and protocol sections.

### DIFF
--- a/client/components/application-summary.js
+++ b/client/components/application-summary.js
@@ -149,7 +149,7 @@ class ApplicationSummary extends React.Component {
                 {
                   subsections.map(key => {
                     const subsection = section.subsections[key];
-                    let fields = Object.values(this.props.fieldsBySection[key] || []);
+                    const fields = Object.values(this.props.fieldsBySection[key] || []);
                     if(subsection.repeats) {
                       fields.push(subsection.repeats);
                     }
--- a/client/components/application-summary.js
+++ b/client/components/application-summary.js
@@ -150,7 +150,9 @@ class ApplicationSummary extends React.Component {
                   subsections.map(key => {
                     const subsection = section.subsections[key];
                     let fields = Object.values(this.props.fieldsBySection[key] || []);
-                    if(subsection.repeats) fields.push(subsection.repeats);
+                    if(subsection.repeats) {
+                      fields.push(subsection.repeats);
+                    }
                     return <tr key={key}>
                       <td><Link to={`/${key}`}>{ subsection.title }</Link></td>
                       <td className="controls">
--- a/client/components/application-summary.js
+++ b/client/components/application-summary.js
@@ -13,11 +13,11 @@ import flatten from 'lodash/flatten';
 import { Button } from '@ukhomeoffice/react-components';
 
 import { INCOMPLETE, PARTIALLY_COMPLETE, COMPLETE } from '../constants/completeness';
-import { LATEST, GRANTED } from '../constants/change';
 import schema from '../schema'
 import { flattenReveals, getNewComments } from '../helpers';
 
 import NewComments from './new-comments';
+import ChangedBadge from './changed-badge';
 
 const getFields = subsection => {
   if (!subsection.repeats && subsection.fields && subsection.fields.length) {
@@ -38,10 +38,6 @@ const mapStateToProps = ({
     showComments,
     user,
     showConditions
-  },
-  changes: {
-    latest,
-    granted
   }
 }) => {
   const fieldsBySection = Object.values(schema[schemaVersion]).map(section => section.subsections).reduce((obj, subsections) => {
@@ -58,9 +54,7 @@ const mapStateToProps = ({
     fieldsBySection: omit(fieldsBySection, 'protocols'),
     legacy: schemaVersion === 0,
     values: project,
-    sections: schema[schemaVersion],
-    latest,
-    granted
+    sections: schema[schemaVersion]
   };
 }
 
@@ -112,27 +106,6 @@ class ApplicationSummary extends React.Component {
     }
   }
 
-  changed = (key, subsection) => {
-    const fields = this.props.fieldsBySection[key] || [];
-    if (this.props.latest.includes(key) || this.props.latest.some(k => fields.includes(k)) || this.props.latest.includes(subsection.repeats)) {
-      return LATEST;
-    }
-    else if (this.props.granted.includes(key) || this.props.granted.some(k => fields.includes(k) || this.props.granted.includes(subsection.repeats))) {
-      return GRANTED;
-    }
-  }
-
-  changedBadge = change => {
-    switch (change) {
-      case LATEST:
-        return <span className="badge changed">changed</span>;
-      case GRANTED:
-        return <span className="badge">amended</span>;
-      default:
-        return null;
-    }
-  }
-
   sectionVisible = section => {
     return !section.show || section.show(this.props.values);
   }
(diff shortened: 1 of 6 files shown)


client/helpers/index.js
```diff
@@ -75,3 +76,12 @@ export const flattenReveals = (fields, values) => {
     ])
   }, []);
 };
+
+export const changed = (fields, latest, granted) => {
+  if(fields.some(k=> latest.some(l=> l.includes(k)))) {
```
joefitter: space after `if`, space after `k`, space after `l` - lint should be picking this up 🤔

lennym:
What was the cause of the bug and how does your code fix it?

client/pages/sections/protocols/sections.js
```diff
@@ -112,25 +113,46 @@ const sortGranted = (sections, isGranted) => (a, b) => {
...
+        <Fragment key={section}>
+        {
+          changedBadge(changed(sections[section].fields.map(field => field.name), props.latest, props.granted))
+        }
         <ExpandingPanel alwaysUpdate={section === 'conditions' || section === 'authorisations'} key={section} title={getTitle(sections[section], newComments, props.values)}>
           {
             getSection(section, { ...props, protocolState, editable, ...sections[section], sectionsLength: size(sections), sectionIndex })
           }
         </ExpandingPanel>
+        </Fragment>
```
lennym: The indentation here is out of whack somehow. There shouldn't be two closing tags at the same level.

client/helpers/index.js
```diff
@@ -75,3 +76,12 @@ export const flattenReveals = (fields, values) => {
...
+    return LATEST;
+  }
+  if(fields.some(k=> granted.some(l=> l.includes(k)))) {
+    return GRANTED;
+  }
+}
```
joefitter: Not sure this is the right place for this "helper" - it is only used once and feels like it should be just defined where it is used - I would suggest creating a <Changed /> or <HasChanged /> or something component and using this wherever we show a changed label

client/components/changed-badge.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+class ChangedBadge extends React.Component {
+  render() {
+    if (this.props.fields.some( k=> this.props.latest.some( l=> l.includes(k)))) {
```
joefitter: spacing still off here, this line should be
```js
if (this.props.fields.some(k => this.props.latest.some(l => l.includes(k)))) {
```

client/components/changed-badge.js
```diff
@@ -0,0 +1,15 @@
...
+      return <span className="badge changed">changed</span>;
+    }
+    if (this.props.fields.some( k=> this.props.granted.some( l=> l.includes(k)))) {
+      return <span className="badge">amended</span>;
+    }
+    return null;
+  }
+}
+
+export default ChangedBadge;
```
joefitter: This file doesn't need a fully fledged React.Component - it doesn't have any lifecycle hooks etc so a plain old function would be better in this case.

client/components/changed-badge.js
joefitter: also what is `k`? Why not `f`, or even better, `field`

client/pages/sections/protocols/sections.js
```diff
@@ -139,7 +148,9 @@ const mapStateToProps = ({
   // show all sections for legacy
   sections: isGranted && schemaVersion !== 0
     ? pickBy(sections, section => section.granted)
-    : sections
+    : sections,
+  latest,
+  granted
```
joefitter: this file no longer cares about latest/granted - why don't we connect them in the new component you made?

client/pages/sections/protocols/protocol-sections.js
```diff
@@ -46,7 +47,9 @@ class ProtocolSections extends PureComponent {
       editable,
       newComments,
       readonly,
-      schemaVersion
+      schemaVersion,
+      latest,
+      granted
```
joefitter: same here, we should be getting these props in the downstream component - `ChangedBadge`

client/components/changed-badge.js
```diff
@@ -1,15 +1,16 @@
...
-class ChangedBadge extends React.Component {
-  render() {
-    if (this.props.fields.some( k=> this.props.latest.some( l=> l.includes(k)))) {
-      return <span className="badge changed">changed</span>;
-    }
-    if (this.props.fields.some( k=> this.props.granted.some( l=> l.includes(k)))) {
-      return <span className="badge">amended</span>;
-    }
-    return null;
+const ChangedBadge = ({ fields, latest, granted }) => {
```
joefitter: I would probably default fields to an empty array here, as it will throw an error if undefined for any reason

joefitter:
@lvnhmd This look much better - one thing I would say is you only seem to be using the new component in protocols - can it be used on the application summary too?

lvnhmd:
> @lvnhmd This look much better - one thing I would say is you only seem to be using the new component in protocols - can it be used on the application summary too?

@joefitter - it most certainly can and it is now

client/components/changed-badge.js
joefitter: BUMP

client/components/changed-badge.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
+const ChangedBadge = ({ fields = [], latest = [], granted = [] }) => {
+  if (fields.some(field => latest.some(change => change.includes(field)))) {
```
joefitter: i didn't actually look at this code in great detail, but I think you want `_.intersection(fields, changes).length`

lennym: I'm totally relaxed about whether or not we leave the "double-some" or use `intersection` (it's broadly the same code under the hood either way), but I have strong feelings about putting this much code _inside_ the condition of an `if`.

It forces the user to understand the code in order to parse the condition, where a descriptively named intermediate value can allow a reader to understand the condition without having to mentally reverse-engineer the code. 

```js
const hasChange = /* whatever implementation you choose */;
if (hasChange) {
  /// stuff
}
```

lennym: #bemorephil

client/components/application-summary.js
```diff
@@ -176,15 +149,15 @@ class ApplicationSummary extends React.Component {
                 {
                   subsections.map(key => {
                     const subsection = section.subsections[key];
+                    let fields = Object.values(this.props.fieldsBySection[key] || []);
+                    if(subsection.repeats) fields.push(subsection.repeats);
```
lennym: Can we always use curly braces for conditionals please?

client/components/application-summary.js
```diff
@@ -176,15 +149,15 @@ class ApplicationSummary extends React.Component {
                 {
                   subsections.map(key => {
                     const subsection = section.subsections[key];
+                    let fields = Object.values(this.props.fieldsBySection[key] || []);
```
lennym: Should be `const` - adding and removing elements of an array doesn't change the value of the `const`.
```suggestion
                    const fields = Object.values(this.props.fieldsBySection[key] || []);
```

client/components/changed-badge.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
+const ChangedBadge = ({ fields = [], latest = [], granted = [] }) => {
+  const hasChangedFromLatest = fields.some(field => latest.some(change => change.includes(field)));
```
lennym: I'd add a check that there are actually _any_ changes before iterating over the fields here. That way we don't end up doing a bunch of iterations that we _know_ are not required.

```suggestion
  const hasChangedFromLatest = latest.length && fields.some(field => latest.some(change => change.includes(field)));
```

client/components/changed-badge.js
```diff
@@ -0,0 +1,18 @@
...
+  if (hasChangedFromLatest) {
+    return <span className="badge changed">changed</span>;
+  }
+  const hasChangedFromGranted = fields.some(field => granted.some(change => change.includes(field)));
```
lennym: As above.
```suggestion
  const hasChangedFromGranted = granted.length && fields.some(field => granted.some(change => change.includes(field)));
```

PR #343, opened 2019-08-12
Improve sync for projects

joefitter:
* added self-calling sync function for projects
* removed the init options for lib, now just takes initialState
* added readonly to state
* added logic for continue to applicationSummary


joefitter:
WIP-ing as breaks drafting

client/actions/projects.js
```diff
@@ -119,26 +120,27 @@ export function updateAndSave(data) {
   };
 }
 
-const shouldSync = (state, fields = []) => {
+const shouldSyncConditions = (state, fields = []) => {
```
lennym: We do a lot of passing `fields` around as an argument, and I think it a) only ever has one value, and b) is only ever actually used in this function.

It might be worth just hard-coding it and not passing it around everywhere.

lennym: Cancel that, it's _also_ used in a `pick` when building the params for the request body.

Having said that, I still think a `const CONDITIONS_FIELDS = ['retrospectiveAssessment', 'conditions'];` at the top of the file might make more sense than passing the argument around everywhere as it still only ever has one value.

PR #344, opened 2019-08-13
removed max-height from images in editor, show errors from editor

joefitter:


lennym:
`throwError` results in a red toaster message, right?

PR #346, opened 2019-08-14
Fix mapping over undefined fields

lennym:
Not all PPL sections have fields, and trying to map over sections which do not have fields throws an error.

### DIFF
--- a/client/pages/sections/protocols/sections.js
+++ b/client/pages/sections/protocols/sections.js
@@ -117,16 +117,20 @@ const ProtocolSections = ({ sections, protocolState, editable, newComments, ...p
   return (
   <Accordion open={getOpenSection(protocolState, editable, sections)} toggleAll={!props.pdf}>
     {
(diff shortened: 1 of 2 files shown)


client/pages/sections/protocols/sections.js
```diff
@@ -117,16 +117,20 @@ const ProtocolSections = ({ sections, protocolState, editable, newComments, ...p
...
-              getSection(section, { ...props, protocolState, editable, ...sections[section], sectionsLength: size(sections), sectionIndex })
-            }
-          </ExpandingPanel>
-        </Fragment>
-      ))
+      Object.keys(sections)
+        .sort(sortGranted(sections, props.isGranted))
+        .filter(section => !sections[section].show || sections[section].show(props))
+        .map((section, sectionIndex) => {
+          const fields = (sections[section].fields || []).map(field => field.name);
```
lennym: This is the only substantive change here - adding ` || []` before the map.

The rest of it is adding whitespace so I could make any sense of the code.

PR #352, opened 2019-08-16
Fix remove button removing the wrong species input

wheelsandcogs:


client/components/species-selector.js
```diff
@@ -48,8 +48,7 @@ class SpeciesSelector extends Component {
       onFieldChange,
       label,
       name,
-      hint,
-      summary
```
wheelsandcogs: summary isn't used

client/components/species-selector.js
```diff
@@ -59,7 +58,7 @@ class SpeciesSelector extends Component {
         }
         {
           map(species, (options, code) => (
-            <details open={this.isOpen(options)}>
+            <details open={this.isOpen(options)} key={code}>
```
wheelsandcogs: fixes a react warning

PR #355, opened 2019-08-20
Legacy species other

joefitter:
* Added component for legacy species selector
* Show text field if other selected
* Added helper to display other species label
* Fixed animal type word download for legacy licenses
* Display correct name in PDF summary tables


client/components/legacy-species-selector.js
```diff
@@ -0,0 +1,39 @@
+import React, { useState, Fragment } from 'react';
+import Field from './field';
+import LEGACY_SPECIES from '../constants/legacy-species';
+
+export default function LegacySpeciesSelector(props) {
+  const otherSpeciesId = LEGACY_SPECIES.find(s => s.label === 'Other species').value;
```
lennym: Given that the id is immutable but the label is not, it's probably better to hard-code the id than the label.

lennym:
```suggestion
  const otherSpeciesId = 28;
```

client/helpers/index.js
```diff
@@ -53,6 +54,15 @@ export const getNewComments = (comments, user) => {
   return pickBy(mapValues(comments, filterNew), filterNew);
 }
 
+export const getLegacySpeciesLabel = species => {
+  const matched = LEGACY_SPECIES.find(s => s.value === species.speciesId);
+  let label = matched && matched.label;
+  if (label === 'Other species') {
+    label = species['other-species-type'];
+  }
```
lennym: Same here as above, it's ok to just use 28 as the id.

client/pages/sections/protocols/legacy-animals.js
```diff
@@ -4,10 +4,10 @@ import Fieldset from '../../../components/fieldset';
 import Repeater from '../../../components/repeater';
 import ReviewFields from '../../../components/review-fields';
 
-const LegacyAnimal = ({ updateItem, values, readonly, prefix, fields, index }) => (
+const LegacyAnimal = ({ updateItem, values, readonly, prefix, fields, index, editable }) => (
   <Fragment>
     {
-      !readonly
+      (!readonly && editable)
```
lennym: Why are these two not equivalent?

Either it's editable _or_ it's read-only surely?

joefitter: readonly is at an application level, editable is at a state level - ie it can be not in edit mode (readonly = false) but check your answers (editable = false)

client/helpers/index.js
```diff
@@ -53,6 +54,15 @@ export const getNewComments = (comments, user) => {
   return pickBy(mapValues(comments, filterNew), filterNew);
 }
 
+export const getLegacySpeciesLabel = species => {
+  const matched = LEGACY_SPECIES.find(s => s.value === species.speciesId);
+  let label = matched && matched.label;
+  if (label === 'Other species') {
```
lennym:
```suggestion
  if (species.speciesId === 28) {
```

client/pages/sections/protocols/legacy-animals.js
lennym: 👍

client/schema/v0/index.js
```diff
@@ -351,16 +350,15 @@ export default {
...
             repeats: 'species',
             fields: [
               {
                 name: 'speciesId',
                 label: 'Animal type',
-                type: 'select',
-                className: 'smaller',
-                options: LEGACY_SPECIES
+                type: 'legacy-species-selector',
+                className: 'smaller'
```
lennym: Options has disappeared because it's now handled inside the component, right?

joefitter: yeah

client/helpers/index.js
joefitter: needs to be a string, I got this

joefitter: bcoz that's what we migrated

PR #360, opened 2019-08-29
Render default duration and licence holder name

lvnhmd:
1.In introductory details section
 *Add PPL holder's name profile link 
2.In .docx add
-Introductory section
*PPL holder's name 
*PPL number
*Primary establishment name
-Establishments page:
*Additional establishment name(s)
*Response to 'Why do you need to use this additional establishments?'
*Response to 'Who will supervise work at this additional establishment?
3.Render default duration

### DIFF
--- a/client/components/download-link/renderers/docx-renderer.js
+++ b/client/components/download-link/renderers/docx-renderer.js
@@ -9,7 +9,7 @@ import SPECIES from '../../../constants/species';
 import { getLegacySpeciesLabel } from '../../../helpers';
 
 export default (application, sections, values, updateImageDimensions) => {
-
+  const insertAtIndex = (arr, index, item) => [...arr.slice(0, index), item, ...arr.slice(index)];
   const numbering = new Numbering();
   const abstract = numbering.createAbstractNumbering();
 
@@ -519,13 +519,6 @@ export default (application, sections, values, updateImageDimensions) => {
     doc.createParagraph(`${years} ${yearsLabel} ${months} ${monthsLabel}`).style('body');
   };
 
-  const renderLicenceHolder = (doc, noSeparator) => {
-    const firstName = application.licenceHolder ? application.licenceHolder.firstName : '';
-    const lastName = application.licenceHolder ? application.licenceHolder.lastName : '';
-    doc.createParagraph('Licence holder').style('Question');
-    renderText(doc, `${firstName} ${lastName}`, noSeparator);
-  };
-
   const renderNull = (doc, noSeparator) => {
     const paragraph = new Paragraph();
     paragraph.style('body');
@@ -632,16 +625,15 @@ export default (application, sections, values, updateImageDimensions) => {
       case 'permissible-purpose':
         renderPermissiblePurpose(doc, field, value, values);
         break;
-      case 'text': {
-        renderText(doc, value, noSeparator);
-        if (field.name === 'title') {
-          renderLicenceHolder(doc);
-        }
-        break;
-      }
+
+      case 'text':
       case 'textarea':
       case 'declaration':
         renderText(doc, value, noSeparator);
+        break;
+
+      case 'holder':
+        renderText(doc, `${value.firstName} ${value.lastName}`, noSeparator);
         break;
 
       case 'texteditor':
@@ -833,6 +825,15 @@ export default (application, sections, values, updateImageDimensions) => {
     .then(() => addImageDimensions(values))
     .then(() => new Document())
     .then(doc => addStyles(doc))
-    .then(doc => renderDocument(doc, sections, values))
+    .then(doc => {
+      const field = {
+        label: 'Licence holder',
+        name: 'holder',
+        type: 'holder'
+      };
+      sections[0].subsections['introduction'].fields = insertAtIndex(sections[0].subsections['introduction'].fields, 1, field);
+      values['holder'] = application.licenceHolder;
+      return renderDocument(doc, sections, values);
+    })
     .then(doc => addPageNumbers(doc))
 }
--- a/client/components/download-link/renderers/docx-renderer.js
+++ b/client/components/download-link/renderers/docx-renderer.js
@@ -510,7 +510,7 @@ export default (application, sections, values, updateImageDimensions) => {
   const renderDuration = (doc, value) => {
     let months = 0;
     let years = 5;
-    if(value) {
+    if (value) {
       months = value.months;
       years = value.years;
     }
@@ -632,15 +632,13 @@ export default (application, sections, values, updateImageDimensions) => {
       case 'permissible-purpose':
         renderPermissiblePurpose(doc, field, value, values);
         break;
-      case 'text': {
-        renderText(doc, value, noSeparator);
-        break;
-      }
+
+      case 'text':
       case 'textarea':
-      case 'declaration': {
+      case 'declaration':
         renderText(doc, value, noSeparator);
         break;
-      }
+
       case 'texteditor':
         renderTextEditor(doc, value, noSeparator);
         break;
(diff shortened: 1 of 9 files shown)


client/components/review-field.js
```diff
@@ -53,6 +53,14 @@ class ReviewField extends React.Component {
       )
     }
 
+    if (this.props.type === 'holder' && this.props.licenceHolder) {
+      return (
+        <a href={`/e/${this.props.establishment.id}/people/${this.props.licenceHolder.id}`}>
+          {`${this.props.licenceHolder.firstName} ${this.props.licenceHolder.lastName}`}
+        </a>
```
lennym: This needs to use a `<Link/>` component because the URL may be different depending on whether it's an internal or external service.

client/components/download-link/renderers/docx-renderer.js
```diff
@@ -578,6 +584,21 @@ export default (application, sections, values, updateImageDimensions) => {
...
+    if (field.type === 'holder') {
+      const firstName = application.licenceHolder ? application.licenceHolder.firstName : '';
+      const lastName = application.licenceHolder ? application.licenceHolder.lastName : '';
+      return renderText(doc, `${firstName} ${lastName}`, noSeparator);
+    }
+
+    if (field.type === 'licenceNumber') {
+      const licenceNumber = application.licenceNumber ? application.licenceNumber : '';
+      return renderText(doc, `${licenceNumber} `, noSeparator);
+    }
```
lennym: Some field types are handled by explicit `if` statements, and others inside a `switch`. Can that be made consistent?

client/components/download-link/renderers/docx-renderer.js
```diff
@@ -493,9 +493,15 @@ export default (application, sections, values, updateImageDimensions) => {
...
 
   const renderDuration = (doc, value) => {
-    let years = value.years === 1 ? 'Year' : 'Years';
-    let months = value.months === 1 ? 'Month' : 'Months';
-    doc.createParagraph(`${value.years} ${years} ${value.months} ${months}`).style('body');
+    let months = 0;
+    let years = 5;
+    if (value) {
+      months = value.months;
+      years = value.years;
```
lennym:
```suggestion
      years = value.years || years;
```

client/components/download-link/renderers/docx-renderer.js
```diff
@@ -493,9 +493,15 @@ export default (application, sections, values, updateImageDimensions) => {
   };
 
   const renderDuration = (doc, value) => {
-    let years = value.years === 1 ? 'Year' : 'Years';
-    let months = value.months === 1 ? 'Month' : 'Months';
-    doc.createParagraph(`${value.years} ${years} ${value.months} ${months}`).style('body');
+    let months = 0;
+    let years = 5;
+    if (value) {
+      months = value.months;
```
lennym:
```suggestion
      months = value.months || months;
```

client/components/download-link/renderers/docx-renderer.js
lennym: These two changes because `value` could be of the form `{ months: 0 }` so is truthy but doesn't contain all values.

client/components/download-link/renderers/docx-renderer.js
```diff
@@ -627,7 +644,7 @@ export default (application, sections, values, updateImageDimensions) => {
           if (step.singular) {
             doc.createParagraph(`${step.singular} ${index + 1}`).heading2();
           }
-          (step.fields || []).filter(f => f.repeats).forEach(field => renderField(doc, field, v, project));
+          (step.fields || []).forEach(field => renderField(doc, field, v, project));
```
lennym: Do we know why has this filter been removed?

client/components/download-link/renderers/docx-renderer.js
wheelsandcogs: I wasn't sure if these were intentionally moved above
```
 if (isUndefined(value) || isNull(value)) {
   return renderNull(doc, noSeparator);
 }
```
to avoid the call to `renderNull`?

client/components/download-link/renderers/docx-renderer.js
wheelsandcogs: Nope.

client/components/review-field.js
wheelsandcogs: `asl-projects` doesn't have `asl-components` as a dep (possibly because of drafting tool?).

I'll add a workaround that doesn't rely on it.

client/components/download-link/renderers/docx-renderer.js
wheelsandcogs: Paging @joefitter

joefitter: :wave

lvnhmd: As per the schema, no fields have repeats and we ended up with the fields themselves not being rendered.

client/components/download-link/renderers/docx-renderer.js
lvnhmd: @wheelsandcogs yes, that is exactly why.  But I agree with @lennym. Replaced ifs with a switch.

client/components/review-field.js
lvnhmd: @wheelsandcogs - have you ? @lennym -  I actually remember having had this conversation with @joefitter https://asl-marvell.slack.com/archives/C988GRGNN/p1566904267018900 , TLDR; use good old fashioned a  but pass it in as different for internal/external.

lennym: The good news is that we have `@as/components` as a dependency now (or at least in one of @wheelsandcogs's PRs), so this can be fixed properly.

wheelsandcogs: https://github.com/UKHomeOffice/asl-projects/pull/367 << here

client/schema/v0/index.js
```diff
@@ -20,6 +20,11 @@ export default {
             name: 'title',
             type: 'text'
           },
+          {
+            name: 'holder',
+            label: 'Licence holder',
+            type: 'holder'
+          },
```
lennym: Does this cause any side-effects when editing a legacy project?

lvnhmd: @lennym - ugh, It does ...

client/components/review-fields.js
```diff
@@ -42,6 +42,8 @@ const ReviewFields = ({ fields, values = {}, onEdit, editLink, project, prefix =
                 noComments={ noComments }
                 altLabels={altLabels}
                 readonly={readonly}
+                licenceHolder={licenceHolder}
+                establishment={establishment}
```
lennym: This looks like it's going to put the licence holder information into every single field, which seems weird.

I'm wary of making these kind of changes to `<ReviewFields/>` and `<ReviewField/>` which are used in a lot of different places in order to deliver putting one piece of content onto one page.

I think adding the licence holder directly to the `legacy-introduction-review.js` file without going via these extra components would be a cleaner solution.

lvnhmd: @lennym - I agree, I struggled with it myself - the reason for passing properties to  ReviewFields is because we want to render the link to user profile below project title, and as it is not a field in schema , if I add the licence holder directly to the legacy-introduction-review.js, I have to render it either above or below all fields https://github.com/UKHomeOffice/asl-projects/blob/92f0f46dbd8e96a1e8204a3081a652edb1ed2006/client/pages/sections/legacy-introduction-review.js#L22, 
would *above* be more acceptable ?

client/components/download-link/renderers/docx-renderer.js
```diff
@@ -611,17 +632,18 @@ export default (application, sections, values, updateImageDimensions) => {
       case 'permissible-purpose':
         renderPermissiblePurpose(doc, field, value, values);
         break;
-
-      case 'text':
+      case 'text': {
+        renderText(doc, value, noSeparator);
+        if (field.name === 'title') {
+          renderLicenceHolder(doc);
+        }
```
lennym: This looks like it's going to end up rendering twice.

I'd be tempted to just pass the licenceHolder as a name string from the point where it is inserted into the values, and not need to do any special processing at all in this component.

lennym: I would give it a custom field type rather than having a special case of `text` though.

client/pages/sections/legacy-introduction-review.js
```diff
@@ -6,47 +6,59 @@ import ReviewFields from '../../components/review-fields';
...
+    const value = {
+      licenceHolder,
+      establishment
+    };
+
+    fields.splice(1, 0, field);
+    values.holder = value;
+  }
+
+  return (
```
lennym: From what I can work out, everything below here is whitespace changes.

lennym:
Note: this will be a breaking change because it changes the schema modules from static data exports to functions.

PR #362, opened 2019-09-06
Fix colspan and rowspans for simple tables

lennym:
The default values for table cells' colspan and rowspan values were being erroneously set to large values. Fix that by defaulting them to 1 where the colspan or rowspan are null.


lennym:
I don't really know how this code works, and I don't have a table with actual rowspans or colspans to test with, so it might not work for those tables. But it was definitely fully broken for simple tables before.

PR #364, opened 2019-09-09
this text cannot be changed so has been removed from the editable aut…

joefitter:
…hs/conditions


joefitter:
the content is hard-coded here https://github.com/UKHomeOffice/asl-projects/blob/master/client/pages/sections/granted/project-summary.js#L93

PR #367, opened 2019-09-12
Use DownloadHeader from asl/components

wheelsandcogs:


client/components/download-header.js
```diff
@@ -1,80 +0,0 @@
...
-  const props = useSelector(selector, shallowEqual);
-  const {
-    id,
-    title,
-    isGranted,
-    drafting,
-  } = props;
-  let { basename = window.location.href } = props;
-
-  basename = basename.replace(/\/edit/, '');
```
wheelsandcogs: I think this is missing from the asl/components version - I guess we need it?

wheelsandcogs: Sorted here https://github.com/UKHomeOffice/asl-projects/pull/367/files#diff-aa416c553720e93e8852480dcca1b545R99

client/project-router.js
```diff
@@ -92,10 +96,12 @@ const ProjectRouter = () => {
     };
   });
 
+  const downloadBasename = basename.replace(/\/edit/, '');
```
lennym: This is a _very_ liberal regex. Does that have the potential to screw future us? Should we at least make sure it appears at the end?

lennym:
```suggestion
  const downloadBasename = basename.replace(/\/edit($|\/)/, '');
```

PR #390, opened 2019-10-10
Allow users to edit their own comments until status change

wheelsandcogs:
Waiting on public-api


client/components/comments/index.js
```diff
@@ -64,10 +75,15 @@ class Comments extends Component {
             )
           }
           {
-            active.map((comment, index) => <Comment index={index + (previous.length || 0)} key={index} field={field} { ...comment } />)
+            active.map((comment, index) => <Comment index={index + (previous.length || 0)} key={index} field={field} editComment={this.editComment} { ...comment } />)
+          }
+          {
+            !!active.length && editing &&
```
lennym:
```suggestion
            editing && commentToEdit &&
```

PR #395, opened 2019-10-14
Update to surgery, anaesthesia and substance administration conditions

wheelsandcogs:
Moves these conditions to be at Protocol summary level in both digital and PDF views.

### DIFF
--- a/client/pages/sections/protocols/protocol-conditions.js
+++ b/client/pages/sections/protocols/protocol-conditions.js
@@ -1,0 +1,51 @@
+import React, { Component, Fragment } from 'react';
+import ReactMarkdown from 'react-markdown';
+import conditions from '../../../constants/protocol-conditions';
+
+class Content extends Component {
+  render() {
+    return (
+      <Fragment>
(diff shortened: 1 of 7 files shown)


client/pages/sections/protocols/protocol-conditions.js
```diff
@@ -0,0 +1,51 @@
...
+        <p>
+          Anaesthesia, surgery, substance administration and withdrawal of fluid conditions relate to each of these protocols.
+        </p>
+
+        {
+          this.props.pdf ? <Content />
+          : (
+            <details>
+              <summary>View conditions</summary>
+              <br />
```
lennym: What's the `<br/>` doing? Seems peculiar.

wheelsandcogs: Copied the html from https://github.com/UKHomeOffice/asl-projects/blob/0339d1cdb2e69b0b78edc9b415786687ac2501bf/client/components/species-selector.js#L80-L88 and didn't want to break any existing styling by faffing with the css.

PR #409, opened 2019-10-23
WIP Add a standalone component for readonly text editors

lennym:
Using this directly in places where we know that the actual editor is not used means we can exclude a number of large modules (editor plugins, jimp) and reduce bundle size on read-only pages.

Note: this has not been thoroughly tested - I'm PR-ing it now to get it off my plate.


client/components/editor/lib/index.js
```diff
@@ -0,0 +1,29 @@
+import { Value } from 'slate';
+import get from 'lodash/get';
+
```
joefitter:
```suggestion
import initialValue from '../initial-value';

```

client/components/editor/readonly.js
```diff
@@ -0,0 +1,46 @@
+import React, { Component, Fragment } from 'react';
+import classnames from 'classnames';
+import { Editor } from 'slate-react';
+
+import { normaliseValue, serialiseValue } from './lib';
+import initialValue from './initial-value';
```
joefitter:
```suggestion
```

client/components/editor/index.js
```diff
@@ -2,16 +2,19 @@ import React, { Component, Fragment } from 'react';
...
 import isEqual from 'lodash/isEqual';
 
 import { throwError } from '../../actions/messages';
 
+import ReadOnlyTextEditor from './readonly';
+
+import { normaliseValue, serialiseValue } from './lib';
+
 import FormatToolbar from './format-toolbar';
 import initialValue from './initial-value'
```
joefitter:
```suggestion
```

joefitter:
U OK HUN?

PR #412, opened 2019-10-23
Preserve paths on edited conditions

lennym:
Where conditions are auto-added but then edited by an inspector the `path` attribute was being stripped, which resulted in being unable to restore the value from the inpector's side.

Ensure that when the criteria for custom conditions are changed that the paths are preserved.

### DIFF
--- a/client/helpers/index.js
+++ b/client/helpers/index.js
@@ -9,32 +9,34 @@ import LEGACY_SPECIES from '../constants/legacy-species';
 export const formatDate = (date, format) => (date ? dateFormatter(date, format) : '-');
 
 export const getConditions = (values, conditions, project) => {
-  const customConditions = (values.conditions || [])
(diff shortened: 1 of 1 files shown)


client/helpers/index.js
```diff
@@ -9,32 +9,34 @@ import LEGACY_SPECIES from '../constants/legacy-species';
...
 
-  conditions.forEach(condition => {
-    const customCondition = customConditions.find(c => c.key === condition.key);
+  const newConditions = conditions.map(condition => {
+    const editedCondition = editedConditions.find(c => c.key === condition.key);
     const path = `${condition.key}.versions.${condition.versions.length - 1}`;
-    if (customCondition) {
-      newConditions.push({
-        ...customCondition,
-        path: condition.path
```
lennym: `condition.path` is not a thing, so is removed. This is the root cause of the issue - everything else is me reformatting and tidying up.

PR #426, opened 2019-10-30
WIP Bugfix - code of practice is a condition

joefitter:


joefitter:
Awaiting confirmation

PR #443, opened 2019-11-01
added logic to hide irrelevant questions on granted licence

joefitter:


lennym:
For my benefit, this story is to not display any content at all for certain questions on the granted licence if answered "No"?

joefitter:
Yes, the ticket says to remove them from granted views if "no", but it seems like it doesn't make sense to leave them if omitted, so I added a falsy check

PR #452, opened 2019-11-04
Bugfix - diff component for not-rte

joefitter:
* Abstracted species selector rendering logic
* Updated diff component to handle species
* Updated to handle permissible purposes
* Handle simple lists


lennym:
Does this need the corresponding middleware change to go at the same time?

joefitter:
for it to work properly, without that the lists provided will be wrong. But if this is merged and bumped in pages then it will go at the same time

joefitter:
https://github.com/UKHomeOffice/asl-pages/pull/583

PR #453, opened 2019-11-04
Bugfix - inform users of errors

joefitter:
previously we were swallowing errors in fields in a promise chain - inform users of these errors


lennym:
Does this give a red banner, or the inline error component?

lennym:
Also

joefitter:
You get the read "toaster" message

lennym:
How difficult would it be to convert it to the error component? This is good, but it's a lot more useful when we actually get bug reports to have more information.

PR #457, opened 2019-11-08
Don't display retrospective assessment as an additional condition

wheelsandcogs:


client/components/conditions/index.js
```diff
@@ -44,19 +42,11 @@ function ConditionsPage({
...
-            <h2>Additional condition 1</h2>
-            <RetrospectiveAssessment />
-          </Fragment>
-        )
-      }
       <Conditions
         scope="project"
         type={type}
         conditions={conditions}
-        startIndex={showRA ? 1 : 0}
```
lennym: It's highly likely that this is no longer required. If it's not then we should remove the prop completely.

PR #460, opened 2019-11-13
Move conditions eval logic

joefitter:
moved logic to debounced sync to prevent evaluating conditions on every keystroke

### DIFF
--- a/client/actions/projects.js
+++ b/client/actions/projects.js
@@ -9,6 +9,7 @@ import * as types from './types';
 import database from '../database';
 import { showMessage, throwError } from './messages';
 import sendMessage from './messaging';
+import { getConditions } from '../helpers';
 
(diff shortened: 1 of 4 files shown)


lennym:
I'd like to get some test coverage for _at least one_ general and protocol level condition in place before merging this.

client/actions/projects.js
```diff
@@ -242,17 +256,26 @@ const syncProject = (dispatch, getState) => {
 
   dispatch(isSyncing());
 
-  const patch = diff(state.savedProject, state.project)
-    .map(d => d.kind === 'edit' ? omit(d, 'lhs') : d);
+  const project = state.application.schemaVersion === 1
```
lennym:
```suggestion
  const project = state.application.schemaVersion > 0
```

PR #466, opened 2019-11-25
Bugfix - hide conditional steps

joefitter:
* If a step is conditional, all fields in the step should be omitted if the condition is fasly

### DIFF
--- a/client/components/static-section.js
+++ b/client/components/static-section.js
@@ -35,7 +35,7 @@ const StaticSection = ({ section, project, fields = [], isGranted, subsection = 
 const mapStateToProps = ({ project, application: { isGranted, schemaVersion } }, { section }) => {
   const fields = flatten([
     ...(section.fields || []),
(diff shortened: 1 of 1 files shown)


client/components/static-section.js
```diff
@@ -35,7 +35,7 @@ const StaticSection = ({ section, project, fields = [], isGranted, subsection =
 const mapStateToProps = ({ project, application: { isGranted, schemaVersion } }, { section }) => {
   const fields = flatten([
     ...(section.fields || []),
-    ...((section.steps || []).map(step => step.fields) || [])
+    ...((section.steps || []).filter(step => !step.show || step.show(project)).map(step => step.fields) || [])
```
lennym:
```suggestion
    ...(section.steps || []).filter(step => !step.show || step.show(project)).map(step => step.fields)
```

Out of scope I know, but also _totally_ redundant so better of removed.

PR #471, opened 2019-11-25
Bugfix - fate of animals hierarchy

joefitter:
* added preserveHeirarchy option to reveal fields - this prevents children being flattened
* render children in scope

### DIFF
--- a/client/components/review-field.js
+++ b/client/components/review-field.js
@@ -5,6 +5,7 @@ import TextEditor from './editor';
 import speciesOptions from '../constants/species';
 import { getLegacySpeciesLabel, mapSpecies } from '../helpers';
 
+import castArray from 'lodash/castArray';
 import flatten from 'lodash/flatten';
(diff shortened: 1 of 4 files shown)


client/components/review-field.js
```diff
@@ -126,11 +128,34 @@ class ReviewField extends React.Component {
...
         <ul>
           {
             value.filter(v => options ? options.find(o => o.value === v) : true).map(value => (
-              <li key={value}>{getValue(value)}</li>
+              <li key={value}>
+                {
+                  getValue(value)
+                }
+                {
+                  this.props.preserveHeirarchy && getChildren(value)
```
lennym: What happens to these children if `preserveHierarchy` is false?

lennym: Looks like it get's caught by `flattenReveals`

joefitter: yeah that, default is to flatten them to top level fields

PR #474, opened 2019-12-03
Post errors

joefitter:
* Added postError action
* Call with error details in ErrorBoundary

needs https://github.com/UKHomeOffice/asl-service/pull/170 to be bumped at same time

### DIFF
--- a/client/actions/messaging.js
+++ b/client/actions/messaging.js
@@ -11,7 +11,7 @@ export function postError({ error, info }) {
     }
   };
 
-  return fetch('/error', params).response;
+  return fetch('/error', params).response.catch(() => {});
 }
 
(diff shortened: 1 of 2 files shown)


client/actions/messaging.js
```diff
@@ -1,6 +1,20 @@
...
+    method: 'POST',
+    credentials: 'include',
+    json: {
+      message: error.message,
+      stack: error.stack,
+      ...info
+    }
+  };
+
+  return fetch('/error', params).response;
```
lennym:
```suggestion
  return fetch('/error', params).response.catch(() => {});
```

lennym: This shouldn't be able to throw more errors.

PR #476, opened 2019-12-03
Deprecate drafting tool

lennym:
Remove ability to create new projects and update shutdown date.


joefitter:
a pleasing amount of red

PR #480, opened 2019-12-05
Restrict image embeds to PNG and JPG files

lennym:
TIFFRAGE!


wheelsandcogs:
What about SVG?

lennym:
SVGs _currently_ fail silently because Jimp doesn't know how to resample them. This actually provides a better experience for someone trying to upload an SVG.

PR #493, opened 2019-12-19
Bugfix - dont try and render changed badge if no changes

joefitter:
* ChangedBadge assumes the redux store has been initialised, it does not when transferring a PPL.
* Check there are changes before trying to render

### DIFF
--- a/client/components/review.js
+++ b/client/components/review.js
@@ -17,23 +17,34 @@ class Review extends React.Component {
 
   render() {
     const { label } = this.props.altLabels ? this.props.alt : this.props;
-    const { hint, isGranted, showGrantedLabel = true, review } = this.props;
+    const {
+      hint,
+      isGranted,
+      showGrantedLabel = true,
+      review,
+      changedFromLatest,
+      changedFromGranted
+    } = this.props;
     return (
       <div className="review">
         {
           (!isGranted || showGrantedLabel) && <h3>{review || label}</h3>
         }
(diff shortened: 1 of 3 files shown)


client/components/review.js
```diff
@@ -17,23 +17,34 @@ class Review extends React.Component {
...
           (!isGranted || showGrantedLabel) && <h3>{review || label}</h3>
         }
-        <ChangedBadge
-          changedFromLatest={this.props.changedFromLatest}
-          changedFromGranted={this.props.changedFromGranted}
-          protocolId={this.props.protocolId}
-        />
         {
-          this.props.readonly && (this.props.changedFromLatest || this.props.changedFromGranted) && (
+          (changedFromLatest || changedFromGranted) && (
```
lennym: Should probably do a `return null` in the `ChangedBadge` component too if the props are falsy.

No objection to having this as well as a belt'n'braces, but putting it in the component feels generally safer.

client/components/review.js
```diff
@@ -17,23 +17,34 @@ class Review extends React.Component {
...
+            <ChangedBadge
+              changedFromLatest={changedFromLatest}
+              changedFromGranted={changedFromGranted}
+              protocolId={this.props.protocolId}
+            />
+          )
+        }
+        {
+          this.props.readonly && (changedFromLatest || changedFromGranted) && (
```
lennym: Same.

PR #494, opened 2019-12-20
Bugfix - permissible purposes

joefitter:
* Added permissible purposes to PDF and granted view for new PPLs
* Updated legacy to use permissible purposes field type to show answers in context

### DIFF
--- a/client/schema/v0/index.js
+++ b/client/schema/v0/index.js
@@ -205,7 +205,7 @@ export default () => ({
             label: 'Which purposes apply to your project?',
             type: 'permissible-purpose',
             className: 'smaller',
-            preserveHeirarchy: true,
+            preserveHierarchy: true,
             options: [
               {
                 label: '(a) Basic research',
--- a/client/schema/v0/index.js
+++ b/client/schema/v0/index.js
@@ -203,8 +203,9 @@ export default () => ({
           {
             name: 'purpose',
             label: 'Which purposes apply to your project?',
-            type: 'checkbox',
+            type: 'permissible-purpose',
             className: 'smaller',
+            preserveHeirarchy: true,
             options: [
               {
                 label: '(a) Basic research',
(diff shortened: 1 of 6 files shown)


views/pdf/v1/index.jsx
```diff
@@ -1,28 +1,28 @@
...
-    <StandardConditions />
-  </Fragment>
-)
-
-const mapStateToProps = ({ project }) => ({ project });
-
-export default connect(mapStateToProps)(Modern);
+export default function PDF() {
+  const project = useSelector(state => state.project);
+  const sections = Object.values(getGrantedSubsections(1)).filter(s => !s.granted.show || s.granted.show(project)).sort((a, b) => a.granted.order - b.granted.order);
```
lennym:
```suggestion
  const sections = Object.values(getGrantedSubsections(1))
    .filter(s => !s.granted.show || s.granted.show(project))
    .sort((a, b) => a.granted.order - b.granted.order);
```

lennym: `s => !s.granted.show` looks like it's going to include if `show: false` which feels like the opposite of what we want.

Perhaps `show === undefined`?

client/schema/v0/index.js
```diff
@@ -203,8 +203,9 @@ export default () => ({
           {
             name: 'purpose',
             label: 'Which purposes apply to your project?',
-            type: 'checkbox',
+            type: 'permissible-purpose',
             className: 'smaller',
+            preserveHeirarchy: true,
```
lennym: How did this ever get through?!

`preserveHierarchy` damnit!

views/pdf/v1/index.jsx
joefitter: not sure, I didn't change this

client/schema/v0/index.js
joefitter: HEHE

views/pdf/v1/index.jsx
joefitter: I _think_ show is _always_ a function or undefined

PR #495, opened 2019-12-20
Bugfix - diff window

joefitter:
* These values are false on render as they aren't calculated until later


client/components/review.js
```diff
@@ -25,6 +25,7 @@ class Review extends React.Component {
       changedFromLatest,
       changedFromGranted
     } = this.props;
+    console.log(this.props.readonly && (changedFromLatest || changedFromGranted))
```
joefitter:
```suggestion
```

PR #499, opened 2020-01-24
Add ability to edit retrospective assessment condition

lennym:
ASRU should be able to add an RA condition if one does not apply, and should see if one has been automatically applied, but it should not appear under the conditions section when viewing the granted licence because RA is already exposed in the "Project summary" section.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,1 +1,2 @@
 public
+test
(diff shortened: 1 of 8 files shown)


lennym:
WIP because the reason I pulled the RA function out into its own file was to add some unit test coverage to it.

PR #503, opened 2020-01-31
Fix for duplicated comments across protocols

wheelsandcogs:


client/pages/sections/protocols/section.js
```diff
@@ -37,8 +37,8 @@ class Section extends PureComponent {
               <ReviewFields
                 fields={fields}
                 values={values}
-                prefix={this.props.prefix}
-                editLink={`0#${this.props.prefix}`}
+                prefix={prefix}
+                editLink={`0#${prefix}`}
```
wheelsandcogs: Not strictly necessary for the fix, but seemed weird to use the full props path when we've already used the destructured `prefix` above.

client/components/comments/index.js
```diff
@@ -94,12 +94,21 @@ class Comments extends Component {
...
-  if (name !== field && field.match(/^protocols\.[a-f0-9-]+\.item\./)) {
-    allComments = [].concat(comments[name]).concat(comments[field]).filter(Boolean);
+  if (name !== field) {
+    // whitelist the fields in legacy protocol animal questions
+    if (field.match(/^protocols\.[a-f0-9-]+\.item\./)) {
+      allComments = [].concat(comments[name]).concat(comments[field]).filter(Boolean);
+    }
+
+    // whitelist the fields nested under the fate section of protocols
+    if (field.match(/^protocols\.[a-f0-9-]+\.killed|killing-method|method-and-justification|continued-use-relevant-project/)) {
```
lennym: Does this need a `$` at the end of the regex to avoid matching partial field names?

wheelsandcogs: It's already limited to just the protocol fields, but sure.

lennym:
```suggestion
    if (field.match(/^protocols\.[a-f0-9-]+\.(killed|killing-method|method-and-justification|continued-use-relevant-project)$/)) {
```

lennym: I'm not 100% confident about regex structure and order-of-operations but I added `(...)` to make sure the `|` were explicitly scoped as well.

PR #512, opened 2020-02-14
Added new view for granted legacy continuation

joefitter:
* Added new legacyGranted setting - isGranted is false for legacy
* Added continuation component with alternate label

### DIFF
--- a/client/components/review-field.js
+++ b/client/components/review-field.js
@@ -56,6 +56,7 @@ class ReviewField extends React.Component {
       )
     }
 
+    // we only want to render this component if viewing a granted licence, if not then fallback to default radio behaviour
     if ((this.props.legacyGranted || this.props.isGranted) && (this.props.name === 'continuation' && !isUndefined(value)))  {
       return (
         <dl className="inline">
--- a/client/components/review-field.js
+++ b/client/components/review-field.js
@@ -57,16 +57,14 @@ class ReviewField extends React.Component {
     }
 
     if ((this.props.legacyGranted || this.props.isGranted) && (this.props.name === 'continuation' && !isUndefined(value)))  {
-      return value
-        ? (
-          <dl className="inline">
-            <dt>From the licence</dt>
-            <dd>{this.props.project['continuation-licence-number']}</dd>
-            <dt>Expiring on</dt>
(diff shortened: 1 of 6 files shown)


client/pages/sections/legacy-introduction-review.js
```diff
@@ -38,8 +39,17 @@ const LegacyIntroduction = ({ fields, project, values, pdf, readonly, title }) =
       )}
       <ReviewFields
         values={values}
-        fields={fields}
+        fields={fields.filter(f => f.name !== 'continuation')}
       />
+    {
+      values.continuation && (
+        <Review
```
lennym: It seems like we've put a load of extra logic into the `<Review/>` component to handle the special case where the field is `continuation`, and it be much cleaner to leave the review component untouched and just put the bespoke content here.

joefitter: yes you're right, it _was_ just included in the `<ReviewFields />` component, but now we've called it out separately it would make sense to do that. I was totally out of fucks by that point

client/components/review-field.js
```diff
@@ -56,6 +56,17 @@ class ReviewField extends React.Component {
       )
     }
 
+    if ((this.props.legacyGranted || this.props.isGranted) && (this.props.name === 'continuation' && !isUndefined(value)))  {
```
lennym: Should the logic here around being a granted version sit outside this component?

As it stands, if it is not a granted version then the component will render `<p>No answer provided</p>` which feels incorrect to me.

client/components/review-field.js
```diff
@@ -56,6 +56,17 @@ class ReviewField extends React.Component {
       )
     }
```
joefitter:
```suggestion

      // we only want to render this component if viewing a granted licence, if not then fallback to default radio behaviour
```

client/pages/sections/legacy-introduction-review.js
```diff
@@ -38,8 +41,18 @@ const LegacyIntroduction = ({ fields, project, values, pdf, readonly, title }) =
...
+          {...continuationField}
+          label={continuationField.grantedLabel}
+          value={values.continuation}
+          values={values}
+        />
+      )
+    }
```
lennym: The indentation has gone a bit wonky here.

PR #513, opened 2020-02-14
Disallow links in comments

joefitter:
* Added custom renderer for links user authored content
* Replaced ReactMarkdown with LinklessMarkdown

### DIFF
--- a/client/components/comments/comment.js
+++ b/client/components/comments/comment.js
@@ -2,7 +2,8 @@ import React, { Fragment } from 'react';
 import { connect } from 'react-redux';
 import classnames from 'classnames';
 import format from 'date-fns/format';
-import { Button, Markdown } from '@ukhomeoffice/react-components';
+import { Button } from '@ukhomeoffice/react-components';
+import { Markdown } from '@asl/components';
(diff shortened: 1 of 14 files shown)


lennym:
Comments from slack for posterity:

Can we remove the "linkless" from the markdown component and move the code into asl/components?

The logic being that there's a lot of non-projects places we also render markdown where the same logic will want to apply.

Remove linkless to avoid any future issues if we add more criteria to our markdown rendering.

joefitter:
WIP until https://github.com/UKHomeOffice/asl-components/pull/116

client/components/comments/comment.js
```diff
@@ -2,8 +2,7 @@ import React, { Fragment } from 'react';
 import { connect } from 'react-redux';
 import classnames from 'classnames';
 import format from 'date-fns/format';
-import ReactMarkdown from 'react-markdown';
-import { Button } from '@ukhomeoffice/react-components';
+import { Button, Markdown } from '@ukhomeoffice/react-components';
```
lennym: Fairly sure that's not what we want.

PR #514, opened 2020-02-17
Licence status should be version status not project status

wheelsandcogs:


wheelsandcogs:
This change is included in https://github.com/UKHomeOffice/asl-pages/pull/705

PR #516, opened 2020-02-17
Handle special cases in Review component

lennym:
Allow implementations of the Review component to pass children in to override the default ReviewField behaviour to replay values. This allows special case renderings to be handled as one-off implementations rather than bundling everything into the Review component.

Implement this for project continuation on legacy licences.

### DIFF
--- a/client/pages/sections/legacy-introduction-review.js
+++ b/client/pages/sections/legacy-introduction-review.js
@@ -50,7 +50,14 @@ const LegacyIntroduction = ({ fields, project, values, pdf, readonly, title }) =
           label={continuationField.grantedLabel}
           value={values.continuation}
           values={values}
-        />
+        >
+          <dl className="inline">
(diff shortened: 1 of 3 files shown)


client/pages/sections/legacy-introduction-review.js
```diff
@@ -50,7 +50,14 @@ const LegacyIntroduction = ({ fields, project, values, pdf, readonly, title }) =
...
+        >
+          <dl className="inline">
+            <dt>From the licence</dt>
+            <dd>{values['continuation-licence-number']}</dd>
+            <dt>Expiring on</dt>
+            <dd>{values['continuation-expiry-date'] && formatDate(values['continuation-expiry-date'], DATE_FORMAT.long)}</dd>
+          </dl>
+        </Review>
       )
     }
```
lennym: I noticed the indentation has gone a bit wonky here.

PR #517, opened 2020-02-19
Fix bug: expired / revoked return false for isGranted

wheelsandcogs:


client/project-router.js
```diff
@@ -108,7 +108,7 @@ const ProjectRouter = () => {
           title={version.title || 'Untitled project'}
           subtitle="Project licence"
           basename={downloadBasename}
-          licenceStatus={isGranted ? project.status : 'inactive'}
+          licenceStatus={project.status === 'active' && !isGranted ? 'inactive' : project.status }
```
lennym: I _think_ this is what you mean, but it's not obvious from the code as it is.

```suggestion
          licenceStatus={ (project.status === 'active' && !isGranted) ? 'inactive' : project.status }
```

PR #529, opened 2020-03-05
Don't filter out deleted protocols when editing

lennym:
This results in the entire set of protocols being shown as part of the diff and massive PPL payloads.

Instead we'll let the client keep the list, and then filter deleted protocols from data at the point of granting.


lennym:
WIP because we still need to filter them out if it's in readonly.

PR #530, opened 2020-03-09
Bugfix - nested lists

wheelsandcogs:


joefitter:
oops, I named by branch the same as yours, then force pushed over your commit. sorrybox

PR #541, opened 2020-03-31
Update wording of marmoset condition

wheelsandcogs:


lennym:
Should probably clarify if this should be a new version of the condition or not.

lennym:
Marked as WIP to prevent accidental merge while we wait for a decision around versioning.

PR #543, opened 2020-04-01
Map step level nmbas to project level

wheelsandcogs:


lennym:
This isn't _actually_ WIP is it, it just needs to deploy with the migration?

wheelsandcogs:
It needs to deploy once the migration has been run, it doesn't necessarily need to go in at the same time because we don't remove the old values from the version data.

PR #544, opened 2020-04-03
Bugfix - denest fate

joefitter:
* Removed fate-nts question, bring fate checkbox group to top level


joefitter:
This will need to ship with the corresponding migration - https://github.com/UKHomeOffice/asl-schema/pull/291

PR #545, opened 2020-04-03
Bugfix - update protocol fate killing method

joefitter:
* Replaced killing method checkbox group with yes no radio


joefitter:
This will need to ship with the corresponding migration

PR #546, opened 2020-04-03
stop adding protocol level nmba condition, use top level instead

wheelsandcogs:


client/constants/conditions.js
```diff
@@ -8,16 +8,6 @@ const species = flatten(values(SPECIES));
...
-    nmbas: {
-      include: protocol => some(protocol.steps, step => step.nmbas),
-      type: 'condition',
-      versions: [
-        {
-          title: 'Neuromuscular blocking agents (NMBAs)',
-          content: 'All work under authority of this protocol involving the use of neuromuscular blocking agents must be carried out in accordance with the Home Office guidelines.'
-        }
-      ]
-    },
```
lennym, line 20: Do we know what effect removing this will have on existing PPLs that have the condition?

PR #547, opened 2020-04-07
Don't show transfer-related content on granted versions

lennym:
Skip all the logic about what content to display in what states, and if the licence is granted just return the basic content immediately.

~WIP because I still have some concerns about ASRU users reviewing not-transfers that I want to investigate.~

### DIFF
--- a/client/components/establishment-selector.js
+++ b/client/components/establishment-selector.js
@@ -10,7 +10,21 @@ const revealContent = `To change the primary establishment you must:
 * review any related sections of the application that appear as ‘incomplete'`
 
 export default function EstablishmentSelector({ value, onFieldChange, review, diff, ...props }) {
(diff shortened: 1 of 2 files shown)


client/components/establishment-selector.js
```diff
@@ -10,7 +10,25 @@ const revealContent = `To change the primary establishment you must:
...
-  const { establishments, canTransfer, establishment, transferInProgress, readonly, project } = useSelector(state => state.application, shallowEqual);
+  const {
+    establishments,
+    canTransfer,
+    establishment,
+    transferInProgress,
+    readonly,
+    project,
+    isGranted,
+    legacyGranted
```
lennym, line 21: The two "granted" props are the only things that have been added here, everything else is newlines.

PR #548, opened 2020-04-15
Bugfix - display transfer proj name if user can see it

joefitter:
* Updated to use establishmentName if not defined, fallback to another est

### DIFF
--- a/client/components/establishment-selector.js
+++ b/client/components/establishment-selector.js
@@ -54,6 +54,9 @@ export default function EstablishmentSelector({ value, onFieldChange, review, di
     }
   }, [localValue]);
 
+  const currentEstId = localValue || establishment.id;
+  const establishmentName = (establishments.find(e => e.id === currentEstId) || {}).name || 'Another establishment';
+
   return (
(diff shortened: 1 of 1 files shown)


lennym:
What's the bug?

joefitter:
sorry - great commit message I know. if the transfer-to field wasn't completed, non-licence holders would always see "another establishment" instead of the current establishment

joefitter:
in edit view

lennym:
And this is because `localValue` isn't/wasn't set in a non-editing context?

It would be good to get an integration test covering this as well as there's been a couple of bugs around this component already.

joefitter:
yeah good shout - I'll add some coverage. Previously we had `const localValue = useState(value || establishment.id)` but this was resulting in the useEffect detecting a change and marking the component as changed as soon as the page was loaded

lennym:
> Previously we had `const localValue = useState(value || establishment.id)`

I don't see that in the red bit of the diff

joefitter:
that was a previous bugfix

PR #551, opened 2020-04-20
Feature - warning when removing a linked item from a protocol

joefitter:
* Added helper to show confirm message when a linked objective, establishment or POLE is removed
* Replaced Objective with standard repeater item


client/helpers/index.js
```diff
@@ -145,3 +145,42 @@ export const stripInvalidXmlChars = text => {
   }
   return text.replace(/([^\x09\x0A\x0D\x20-\uD7FF\uE000-\uFFFC\u{10000}-\u{10FFFF}])/ug, '')
 };
+
+// protocol 1, protocol 2 and protocol 3
```
lennym, line 149: Not sure what this comment is for, but it seems to represent neither the arguments nor return value of this function.

client/helpers/index.js
```diff
@@ -145,3 +145,42 @@ export const stripInvalidXmlChars = text => {
...
+
+  // item doesn't appear in any protocols
+  if (!protocols.length) {
+    return true;
+  }
+
+  let message = `Protocols will be affected\n\nRemoving this ${singularName} will also remove it from`;
+
+  // item appears in all protocols
+  if (protocols.length === (project.protocols || []).length) {
```
lennym, line 176: Does this handle deleted protocols?

client/helpers/index.js
joefitter: it's _one_ of the return values - I can expand it a bit

client/helpers/index.js
joefitter: good point, bro###

client/helpers/index.js
lennym: The whole map just results in `1, 2 and 3` though doesn't it? I don't see where "protocol" would be added.

client/pages/sections/repeater/index.js
```diff
@@ -1,27 +1,37 @@
...
-      <Fieldset
-        fields={fields}
-        values={values}
-        prefix={prefix}
-        onFieldChange={(key, value) => updateItem({ [key]: value })}
-      />
-    </div>
-  </Fragment>
-)
+export function Item({ index, fields, values, updateItem, removeItem, length, prefix, singular, confirmRemove }) {
```
lennym, line 10: Can we whack some newlines in here please? It's getting pretty girthy.

client/pages/sections/repeater/index.js
```diff
@@ -1,27 +1,37 @@
...
+  const project = useSelector(state => state.project);
+  function confirmRemoveItem(e) {
+    e.preventDefault();
+    if (confirmRemove(project, values)) {
```
lennym, line 14: This probably wants a check that `confirmRemove` is defined and is a function.

client/helpers/index.js
```diff
@@ -145,3 +145,45 @@ export const stripInvalidXmlChars = text => {
...
+    return window.confirm(`${message} all protocols.`)
+  }
+
+  const protocolText = protocols.length === 1
+    // item appears in a single protocol
+    ? `protocol ${protocols[0].index + 1}`
+    // item appears in many protocols, list them
+    : `protocols ${getProtocolParts(protocols)}`;
+  return window.confirm(`${message} ${protocolText}.`);
+};
```
lennym, line 189: Could we split some of "helpers" out into individual files? Massive dumping ground index.js files are kind of a pain to work with.

Suggest creating `confirm-remove.js` and putting the green bits of this diff into it.

client/helpers/index.js
joefitter: yeah that's what I meant, derp - I'll fix the comment

PR #560, opened 2020-05-05
Feature - protocol summary table

joefitter:
* added summary table page
* added link to protocol views
* Updated existing protocol summary to include expandable row

### DIFF
--- a/assets/scss/projects.scss
+++ b/assets/scss/projects.scss
@@ -29,6 +29,7 @@ html.modal-open, html.modal-open body {
   }
 
   .protocols-summary {
+    // 7th child is the rowspan column appearing right after a non rowspan col
     th:nth-child(7),
     td:nth-child(7) {
       padding-left: 20px;
--- a/assets/scss/projects.scss
+++ b/assets/scss/projects.scss
@@ -28,6 +28,48 @@ html.modal-open, html.modal-open body {
     }
   }
 
+  .protocols-summary {
+    th:nth-child(7),
+    td:nth-child(7) {
+      padding-left: 20px;
+    }
+
+    tr.expandable {
+      cursor: pointer;
+
+      &.hover, &.expanded {
+        background-color: white;
+      }
+
+      &.expanded {
+        td.is-last-row {
+          border-bottom: 0;
+        }
+      }
+    }
+
+    .expanding-container {
+      background-color: white;
+      padding: $govuk-gutter;
+    }
+
+    span.expanding {
+      &:before {
+        content: '►';
+        font-size: 1em;
+        vertical-align: middle;
+        margin-right: 10px;
+      }
+
+      &.expanded {
+        &:before {
+          content: '▼';
+          margin-right: 8px;
+        }
+      }
+    }
+  }
+
   .application-summary {
     h2.section-heading {
       margin-bottom: 10px;
--- a/client/pages/sections/granted/summary-table.js
+++ b/client/pages/sections/granted/summary-table.js
@@ -1,91 +1,168 @@
-import React, { Fragment } from 'react';
+import React, { Fragment, useState } from 'react';
+import classnames from 'classnames';
 import { getLegacySpeciesLabel } from '../../../helpers';
 import { filterSpeciesByActive } from '../protocols/animals';
 
-const SummaryTable = ({ protocols, isLegacy, project }) => (
-  <div className="summary-table">
-    <h3>Summary table</h3>
-    <table>
-      <thead>
-        <tr>
-          <th>No.</th>
-          <th>Protocol</th>
-          <th>Animal types</th>
+function titleCase(str) {
+  return `${str.charAt(0).toUpperCase()}${str.substring(1)}`;
+}
+
+function TableRow({ species, protocol, index, isLegacy, ExpandingRow, expanded, onClick }) {
+  // this is necessary due to :hover css not taking rowspan into account.
+  const [hover, setHover] = useState(false)
+  function onMouseEnter() {
+    setHover(true);
+  }
+
+  function onMouseLeave() {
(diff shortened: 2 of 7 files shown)


client/pages/sections/granted/summary-table.js
```diff
@@ -1,91 +1,168 @@
...
                 {
-                  speciesIndex === 0 && (
-                    <Fragment>
-                      {
-                        !isLegacy && <td rowSpan={species.length || 1}>{ protocol.gaas === true ? 'Yes' : 'No' }</td>
-                      }
-                      <td rowSpan={species.length || 1}>{ protocol.severity }</td>
-                    </Fragment>
-                  )
+                  !isLegacy && <td rowSpan={species.length || 1} className="is-last-row">{ protocol.gaas === true ? 'Yes' : 'No' }</td>
```
lennym, line 63: Given that the majority of the cells in the row are either wrapped in - or contain - an `isLegacy ? ... ` would it be cleaner to do it once, and have a fragment for each option?

i.e.

```
isLegacy ?
  <Fragment>
    <td>...</td>
    <td>...</td>
    <td>...</td>
  </Fragment> :
  <Fragment>
    <td>...</td>
    <td>...</td>
    <td>...</td>
  </Fragment>
```

assets/scss/projects.scss
```diff
@@ -28,6 +28,48 @@ html.modal-open, html.modal-open body {
     }
   }
 
+  .protocols-summary {
+    th:nth-child(7),
+    td:nth-child(7) {
```
lennym, line 33: 🚨 MAGIC NUMBERS 🚨 

Is this actually a `:last-child` or is there something else going on here? At the very least a comment to explain would be beneficial.

assets/scss/projects.scss
```diff
@@ -28,6 +28,48 @@ html.modal-open, html.modal-open body {
...
+
+    tr.expandable {
+      cursor: pointer;
+
+      &.hover, &.expanded {
+        background-color: white;
+      }
+
+      &.expanded {
+        td.is-last-row {
```
lennym, line 45: Why not a `:last-child`?

client/pages/sections/granted/summary-table.js
```diff
@@ -1,91 +1,168 @@
...
-                  {
-                    isLegacy
-                      ? s['life-stages']
-                      : (s['life-stages'] || []).join(', ')
-                    }
-                </td>
-                {
-                  isLegacy && <td>{s['genetically-altered'] === true ? 'Yes' : 'No'}</td>
-                }
+          <td className={classnames({ 'is-last-row': isLastRow })}>
```
lennym, line 49: As mentioned above, using `:last-child` for this would be preferable.

lennym: Oh wait, this isn't _actually_ the last row is it? It's the last _species_. Is that right?

client/pages/sections/granted/summary-table.js
joefitter: probably - none of the isLegacy stuff is new and I didn't want to wander too far out of my lane

assets/scss/projects.scss
joefitter: yeah it sucks - but I couldn't find a way to select the rowspan item which occurs right after a non rowspan. I'll add a comment

assets/scss/projects.scss
joefitter: doesn't work - we want it for all cells in the last row, which can be the first row or the nth row. Because rowspan

client/pages/sections/granted/summary-table.js
joefitter: yes - that - technically it is a row, but there can b n rows per species

assets/scss/projects.scss
```diff
@@ -28,6 +28,48 @@ html.modal-open, html.modal-open body {
     }
   }
 
+  .protocols-summary {
```
joefitter, line 31:
```suggestion
  .protocols-summary {
    // 7th child is the rowspan column appearing right after a non rowspan col
```

client/pages/sections/granted/summary-table.js
joefitter: I'll do the thing.

joefitter: TBF half of them are dependent on legacy, and half not, so refactoring would result in half the rows being repeated. I'm not sure that's preferable

lennym: I'm fairly sure I prefer it, but it's hard to say without seeing it.

lennym: It looks like it's the middle 4 cells in the row are all dependent on legacy-ness, so only include those in the condition, and leave the first two and last cells out and there shouldn't be much repetition.

assets/scss/projects.scss
lennym: I'm still not sure I get this.

`tr:last-child td` would match all the cells in the last row, right?

joefitter: yep, but not the cells in the first row, which are actually the last row cos rolspan

joefitter: also it's not the last child, there are multiple <tr>s per protocol, so it's the last species in a a protocol, I tried with css and couldn't get anything close without a class.

lennym: 🤯

lennym: #shipit then.

joefitter: Screen Shot 2020-05-06 at 11 42 59
VS
Screen Shot 2020-05-06 at 11 43 58

PR #561, opened 2020-05-11
Remove drafting-tool-ness

lennym:
Adds the minimum necessary to run a transient local instance, but with as many references to the drafting tool as I could find removed.


assets/scss/app.scss
```diff
@@ -8,3 +8,7 @@ $assetPath: '/public';
 @import "@ukhomeoffice/frontend-toolkit/assets/styles/components/table";
 
 @import "projects";
+
+.download-header {
+  display: none;
```
lennym, line 13: It was broken, and won't work without a PDF renderer, so just hide it.

assets/scss/app.scss
```diff
@@ -1,5 +1,5 @@
 $assetPath: '/public';
-
+$red: #d4351c;
```
lennym, line 2: I have no idea where we get this from in the main app, but I can't get css to build here without it.

wheelsandcogs: I think it's defined in Components for the diff:
https://github.com/UKHomeOffice/asl-components/blob/fc8aca46fdc3fc2c0fc3bf9562ea669776ffe28f/src/diff-text/index.scss#L1

lennym:
Closing this because I don't think it is going anywhere as it is.

PR #562, opened 2020-05-12
Refactor syncing

lennym:
Use a comparison between the saved state and the local state to determine whether to use an `onbeforeunload` function instead of the `isSyncing` which only tracks if an http request is in progress, and is false if the sync is in a debounced state.

Increase the debounce limits on syncing to reduce the total number of syncs when an application is being actively worked on.

### DIFF
--- a/client/actions/projects.js
+++ b/client/actions/projects.js
@@ -256,6 +256,8 @@ function getProjectWithConditions(project) {
   };
 }
 
+const wait = t => new Promise(resolve => setTimeout(resolve, t));
+
 const syncProject = (dispatch, getState) => {
   const state = getState();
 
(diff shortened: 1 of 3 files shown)


client/project-router.js
```diff
@@ -10,15 +11,27 @@ import ProtocolSummary from './pages/sections/protocols/summary-table';
...
     project,
-    isSyncing,
     basename,
     drafting,
     isGranted,
     legacyGranted
   }
-}) => ({ project, version, isSyncing, basename, drafting, isGranted, legacyGranted });
+}) => {
+  const isSyncing = !readonly && !isEqual(version, savedProject);
```
lennym, line 24: This is the magic here.

client/actions/projects.js
```diff
@@ -370,7 +373,7 @@ export function fetchQuestionVersions(key) {
 
 const debouncedSyncProject = debounce((...args) => {
   return syncProject(...args);
-}, 1000, { maxWait: 5000, leading: true });
+}, 2000, { maxWait: 20000, leading: false });
```
lennym, line 376: This should help reduce a bit of API load too since it throttles down the rate at which changes sync if the application is changing continually.

PR #567, opened 2020-05-19
Feature - Full application PDF

joefitter:
* Added fullApplication option when downloading a pdf
* Updated protocols to include repeating continued header
* added hacky white box to cover the protocol continued header on first page
* only sort protocol sections if granted

### DIFF
--- a/client/components/nts.js
+++ b/client/components/nts.js
@@ -5,19 +5,19 @@ import { Button } from '@ukhomeoffice/react-components';
 export default function NTS({ accepted, onAccept, review, hideTitle }) {
   const isFullApplication = useSelector(state => state.application.isFullApplication);
 
+  if (isFullApplication) {
+    return null;
+  }
+
   return (
     <div className="nts">
       {
         !hideTitle && <h3>Creating your non-technical summary</h3>
       }
-      {
-        !isFullApplication && (
-          <Fragment>
-            <p>Any information you enter on this screen will form part of a non-technical summary (NTS) on GOV.UK, so use everyday language that can be easily understood by members of the public.</p>
-            <p>As an NTS is a public document, do not include any information that could identify any people, locations, or intellectual property related to your project. You will be able to review and edit your NTS as a whole before you send your licence application to the Home Office.</p>
-          </Fragment>
-        )
-      }
+      <Fragment>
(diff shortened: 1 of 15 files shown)


client/components/nts.js
```diff
@@ -1,19 +1,28 @@
...
+  const isFullApplication = useSelector(state => state.application.isFullApplication);
 
-export default NTS;
+  return (
+    <div className="nts">
+      {
+        !hideTitle && <h3>Creating your non-technical summary</h3>
+      }
+      {
+        !isFullApplication && (
```
lennym, line 14: Is there a context we want to render _anything_ in this component is `isFullApplication` is true?

If not then I'd prefer to return an early `null` rather than relying on the other props being set correctly. It feels a little fragile.

client/components/static-section.js
```diff
@@ -2,12 +2,17 @@ import React, { Fragment } from 'react';
...
 
 const StaticSection = ({ section, project, fields = [], isGranted, subsection = false, ...props }) => {
-  const Component = isGranted
+  let Component = isGranted && !props.isFullApplication
     ? (section.granted && section.granted.review) || section.review || ReviewFields
     : section.review || ReviewFields;
 
+  if (props.pdf && section.name === 'protocols') {
+    Component = PDFProtocols
+  }
```
lennym, line 14: This feels like something that might be better done inside the protocol component, as per https://github.com/UKHomeOffice/asl-projects/blob/master/client/pages/sections/granted/protocols.js#L8-L10

The logic gets a bit spread around otherwise.

joefitter: that is a good shout - I didn't even look at that file. so. many. different. versions. of. protocols.

client/components/nts.js
joefitter: yeah I think we can return early here

lennym:
Marking as WIP to quieten the noise until it's ready to go.

client/components/static-section.js
joefitter: That doesn't actually work, as that file is only used if you're not viewing the "full application", which is the case this is covering. This goes back to the problem with no longer having a read only view, and a granted read only view - now there is another type of view which we include in the pdf 🙄

lennym: Not related to the above comment, but I think L8-14 wants a bit of a refactor to make the conditions clearer.

I'd probably pull the logic out into a `getComponent` function so it can examine each case and return early as appropriate.

PR #568, opened 2020-05-22
Move syncing tracking code out of router into separate component

lennym:
Renders an empty div to the page with some classes to indicate sync state to allow for easier tracking in functional tests, and potentially a starting point for future UI as well.

### DIFF
--- a/client/components/sync-handler.js
+++ b/client/components/sync-handler.js
@@ -1,0 +1,61 @@
+import React, { useEffect } from 'react';
+import { useSelector, shallowEqual } from 'react-redux'
+import isEqual from 'lodash/isEqual';
+import classnames from 'classnames';
+
+const selector = ({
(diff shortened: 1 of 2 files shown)


client/components/sync-handler.js
```diff
@@ -0,0 +1,61 @@
...
+    editConditions
+  }
+}) => {
+  // only compare version data if version is in an editable state
+  const editable = !readonly || editConditions;
+  const isSyncing = editable && !isEqual(version, savedProject);
+  return { isSyncing };
+};
+
+const LoadingIndicator = () => {
```
wheelsandcogs, line 20: Can we sync the const name and the file name? :D

lennym: I see what you did there.

They used to be synced, and then I changed one of them.

PR #573, opened 2020-06-05
Dont mutate values

joefitter:
setting values.title to a default value causes project and savedProject to be out of sync. This causes the page to block redirects and refreshing


lennym:
Good catch.

The good news is that it only seems to manifest if you submit a project with no title, so it's wildly unlikely to hit prod. Is that right?

joefitter:
yes, except it's if you have an unnamed protocol.

lennym:
ORITE. The point still stands.

PR #583, opened 2020-06-19
Send version as header to server

lennym:
This allows the local and server versions to be validated for compatibility and incompatible old versions to be prompted to refresh.

### DIFF
--- a/client/actions/projects.js
+++ b/client/actions/projects.js
@@ -232,7 +232,10 @@ const onSyncError = (func, err, dispatch, getState, ...args) => {
   console.error(err);
   dispatch(doneSyncing());
   dispatch(syncError());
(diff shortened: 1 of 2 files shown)


client/actions/projects.js
```diff
@@ -232,7 +232,10 @@ const onSyncError = (func, err, dispatch, getState, ...args) => {
   console.error(err);
   dispatch(doneSyncing());
   dispatch(syncError());
-  dispatch(throwError('Failed to save, trying again'));
+  dispatch(throwError(err.code === 'UPDATE_REQUIRED'
+    ? 'This software has been updated. Please reload your browser to get the latest version.'
```
wheelsandcogs, line 236: Is this the final text?

lennym: It is now.

PR #585, opened 2020-07-01
Handle errors in rich text rendering marginally more gracefully

lennym:
Instead of crashing the entire document, replace the failing node with an error block.

Input (with bold text set to throw an error)
image

Output
image

### DIFF
--- a/client/components/download-link/renderers/docx-renderer.js
+++ b/client/components/download-link/renderers/docx-renderer.js
@@ -106,6 +106,13 @@ export default (application, sections, values, updateImageDimensions) => {
       .quickFormat()
       .font('Helvetica')
       .size(20);
+
+    document.Styles.createParagraphStyle('error', 'Error')
+      .basedOn('Body')
+      .next('Body')
+      .quickFormat()
+      .color('FF0000')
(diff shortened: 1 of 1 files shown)


client/components/download-link/renderers/docx-renderer.js
```diff
@@ -361,7 +368,14 @@ export default (application, sections, values, updateImageDimensions) => {
     }
     const nodes = content.document.nodes;
 
-    nodes.forEach(node => renderNode(doc, node));
+    nodes.forEach(node => {
+      try {
+        renderNode(doc, node);
+      } catch (e) {
+        doc.createParagraph(`There was a problem rendering this content (${node.type})`).style('error');
+        doc.createParagraph(e.stack).style('error');
```
wheelsandcogs, line 376: is this going to dump the stack trace in production?

lennym: Yes.

Given that a) it's super rare that these error, b) there's no actual logging done so this is likely our best shot at debugging, and c) this is a public repo anyway I'm not totally bothered by that fact.

PR #555, opened 2020-04-29
WIP modify the schema to show relevant training questions for training ppls

wheelsandcogs:
Waiting on feedback about specific questions from team.

PR #586, opened 2020-07-09
Training PPLs

lennym:
* Rebases #555 to remove the standalone changes, and preserve only the schema modifications
* Add permissible purpose section to granted PPL outputs

### DIFF
--- a/client/pages/sections/protocols/section.js
+++ b/client/pages/sections/protocols/section.js
@@ -14,7 +14,8 @@ class Section extends PureComponent {
       prefix,
       editable,
       pdf,
-      title
+      title,
+      project
     } = this.props;
 
     return (
@@ -27,7 +28,7 @@ class Section extends PureComponent {
           editable && !values.deleted
             ? (
               <Fieldset
-                fields={fields}
+                fields={fields.filter(f => f.show === undefined || f.show(project))}
                 values={values}
                 prefix={prefix}
                 onFieldChange={onFieldChange}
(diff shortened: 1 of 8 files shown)


client/pages/sections/protocols/section.js
```diff
@@ -27,15 +29,15 @@ class Section extends PureComponent {
...
+                fields={fields.filter(f => f.show === undefined || f.show(project))}
                 values={values}
                 prefix={prefix}
                 onFieldChange={onFieldChange}
               />
             )
             : (
               <ReviewFields
-                fields={fields}
+                fields={fields.filter(f => f.show === undefined || f.show(project))}
```
lennym, line 40: This should probably move inside the component and be handled there, but that's probably a thing for another PR.

client/pages/sections/protocols/section.js
```diff
@@ -48,4 +50,4 @@ class Section extends PureComponent {
   }
 }
 
-export default withRouter(Section);
+export default withRouter(connect(({ project }) => ({ project }))(Section));
```
wheelsandcogs, line 53: holy brackets batman

client/pages/sections/protocols/section.js
lennym: In fact, it looks like it already _is_ handled in `ReviewFields`

https://github.com/UKHomeOffice/asl-projects/blob/master/client/components/review-fields.js#L10-L18

client/pages/sections/protocols/section.js
lennym: Fair. Cleaned up a bit.

PR #601, opened 2020-08-04
Increase the size of RTE icons

mikieet6:


client/components/editor/format-toolbar.js
```diff
@@ -41,7 +41,7 @@ class FormatToolbar extends Component {
           title={tooltip}
         >
           <input type='file' onChange={this.onClickImage} />
-          <Icon icon={icon} />
+          <Icon icon={icon} size={'24'} />
```
lennym, line 44: There's a couple more of these further down the file that will need the same treatment.

PR #604, opened 2020-08-06
Prevent transfer message showing in review mode

lennym:
All PPL applications are showing a "transfer requested" message, even if they're first time applications. This is because the establishment selector component is now used in first-time applications as well as amendments, but did not have the internal logic updated to reflect this.

Return the simple establishment playback early in review, in the same way as for granted licences. Also remove the double playback of establishment from the establishments review section, so it only appears once.


lennym:
Screenshot 2020-08-06 at 09 55 01

PR #612, opened 2020-08-11
Export non-connected instance of ReviewFields

lennym:
This is used in some contexts where the redux store does not exist - specifically in asl-pages for reviewing PPL holder changes.

In those cases we need to be able to pass the `project` prop in directly in order to render conditional fields correctly.


joefitter:
seems legit

PR #616, opened 2020-08-19
Confirm edit if protocols affected

wheelsandcogs:
Adds an edit toggle to the POLE name / Objective title fields that will trigger a warning message if any protocols will be affected by the change.

### DIFF
--- a/client/helpers/confirm-protocols-affected.js
+++ b/client/helpers/confirm-protocols-affected.js
@@ -14,10 +14,11 @@ function getProtocolParts(protocols) {
   }).join('');
 }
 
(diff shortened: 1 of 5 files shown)


client/helpers/confirm-protocols-affected.js
```diff
@@ -14,18 +14,25 @@ function getProtocolParts(protocols) {
   }).join('');
 }
 
-const confirmRemove = (protocolRef, singularName, key) => (project, item) => {
+const confirmProtocolsAffected = (changeType, protocolRef, singularName, key, label) => (project, item) => {
```
joefitter, line 17: I was going to say can we have an object for the first arg? as it's impossible to tell in schema which args are expected in which order

PR #621, opened 2020-08-24
Accessibility fix - editor controls

joefitter:
Added labels to each button, hidden with css


wheelsandcogs:
You probably tried this, but what about adding `aria-label={tooltip}` to the `<button>` rather than text that has to be hidden?

PR #630, opened 2020-10-01
Use diffSentences to compare long passages

lennym:
There is a PPL amendment in prod with a 4,500 word answer that contains a lot of changes. Trying to view the diff of that answer using `diffWords` causes the pages to crash.

Replacing `diffWords` with `diffSentences` for long passages vastly improves the performance of the component.

### DIFF
--- a/client/components/diff-window.js
+++ b/client/components/diff-window.js
@@ -2,7 +2,7 @@ import React, { Fragment } from 'react';
 import { connect } from 'react-redux';
 import classnames from 'classnames';
 import { Value } from 'slate';
-import { diffWords, diffArrays } from 'diff';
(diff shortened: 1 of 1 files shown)


client/components/diff-window.js
```diff
@@ -92,8 +92,11 @@ class DiffWindow extends React.Component {
         } catch (e) {
           return { added: [], removed: [] };
         }
-
-        diffs = diffWords(before.document.text, after.document.text);
+        if (before.document.text.length < 1000 && after.document.text.length < 1000) {
```
wheelsandcogs, line 95: This is 1k chars which is roughly 150 words or about a paragraph - could we maybe give them ~3 paragraphs before switching or is that still too slow?

PR #638, opened 2020-11-12
Bugfix - transfer AA est

joefitter:
* If transfer, use transfer to est for est selector
* Try to lookup est name from assoc ests
* uniq est names in protocol est

### DIFF
--- a/client/helpers/get-locations.js
+++ b/client/helpers/get-locations.js
@@ -1,9 +1,21 @@
-export default function getLocations (project, establishment) {
-  const establishments = (project.establishments || []).filter(e => e.name || e['establishment-name']).map(e => e.name || e['establishment-name']);
+import uniq from 'lodash/uniq';
+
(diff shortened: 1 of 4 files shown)


client/helpers/get-locations.js
```diff
@@ -1,9 +1,21 @@
-export default function getLocations (project, establishment) {
-  const establishments = (project.establishments || []).filter(e => e.name || e['establishment-name']).map(e => e.name || e['establishment-name']);
+import uniq from 'lodash/uniq';
+
+export default function getLocations (project, establishment, establishments) {
+  let est = establishment;
+  if (project.transferToEstablishment) {
+    est = establishments.find(e => e.id === project.transferToEstablishment);
```
lennym, line 6: I'm a little concerned that this may cause issues if an admin is editing a PPL transfer initiated by the PPL holder, since if I understand it correctly, `establishments` in this context is the current user's establishment list.

PR #639, opened 2020-11-13
Bugfix/set est name for aa

joefitter:


client/actions/projects.js
```diff
@@ -376,8 +376,8 @@ const debouncedSyncProject = debounce((...args) => {
 
 export const ajaxSync = props => {
   return (dispatch, getState) => {
-    const { project, application: { establishment } } = getState();
-    const newState = cleanProtocols(project, props, establishment);
+    const { project, application: { establishment, establishments } } = getState();
+    const newState = cleanProtocols(project, props, establishment, establishments);
```
lennym, line 380: I think these changes are redundant now.

joefitter: right you are, missed this part

PR #649, opened 2020-11-25
Sticky headers

wheelsandcogs:
~Waiting on https://github.com/UKHomeOffice/asl-components/pull/177 bump~

### DIFF
--- a/assets/scss/projects.scss
+++ b/assets/scss/projects.scss
@@ -976,6 +976,12 @@ html.modal-open, html.modal-open body {
     }
   }
 
+  .document-header .page-title {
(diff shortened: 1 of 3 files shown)


assets/scss/projects.scss
```diff
@@ -6,7 +6,7 @@ $pink: #ff1493;
 @mixin sticky {
   position: -webkit-sticky;
   position: sticky;
-  top: 0;
+  top: 2 * $govuk-gutter;
```
wheelsandcogs, line 9: make sure side nav leaves room for sticky header.

PR #660, opened 2020-12-03
Give the modal a z-index boost to bring it higher than the header

wheelsandcogs:
~waiting on https://github.com/UKHomeOffice/asl-components/pull/183~

### DIFF
--- a/assets/scss/projects.scss
+++ b/assets/scss/projects.scss
@@ -100,7 +100,7 @@ html.modal-open, html.modal-open body {
   }
 
   .modal {
-    z-index: 100;
+    z-index: 1000;
     position: fixed;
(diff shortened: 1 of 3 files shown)


assets/scss/projects.scss
```diff
@@ -100,7 +100,7 @@ html.modal-open, html.modal-open body {
   }
 
   .modal {
-    z-index: 100;
+    z-index: 1000;
```
joefitter, line 103:
```suggestion
    z-index: 99999999999999999999999999999999;
```

wheelsandcogs: I think the alerts need to be above it ;D

PR #665, opened 2020-12-22
Refactor diff window component

lennym:
This should now load the diff data from the server response instead of comparing locally.


lennym:
This now also includes a rebuild of permissible purpose playback so it maintains the nested lists for sub-purposes of B.

Screenshot 2020-12-22 at 13 57 17

PR #675, opened 2021-01-06
un-nest acclimatise question and remove parent question

wheelsandcogs:


lennym:
[This line](https://github.com/UKHomeOffice/asl-autoproject/blob/master/test/ppl/apply/specs/index.js#L280) will need to be removed also (I would do it now but it will interfere with in-progress deployments)

PR #683, opened 2021-01-22
Move expander into hint text so it can appear between label and radio buttons

wheelsandcogs:


lennym:
This needs a components bump before merging, right?

wheelsandcogs:
> This needs a components bump before merging, right?

Nope, it uses https://github.com/UKHomeOffice/react-components/blob/master/src/forms/radio-group.jsx not `@asl/components/fieldset`, the components change is for pages.

PR #686, opened 2021-02-01
Add missing hint text to application pdf

wheelsandcogs:
and show the missing questions in protocol steps.

### DIFF
--- a/client/pages/sections/protocols/steps.js
+++ b/client/pages/sections/protocols/steps.js
@@ -167,6 +167,8 @@ class Step extends Component {
 
 export default function Steps({ values, prefix, updateItem, editable, ...props }) {
   const isReviewStep = parseInt(useParams().step, 10) === 1;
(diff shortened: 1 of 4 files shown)


client/pages/sections/protocols/steps.js
```diff
@@ -167,6 +167,8 @@ class Step extends Component {
 
 export default function Steps({ values, prefix, updateItem, editable, ...props }) {
   const isReviewStep = parseInt(useParams().step, 10) === 1;
+  const isFullApplicationPdf = props.isFullApplication && props.pdf;
+  const addAnother = !values.deleted && !!editable && !isFullApplicationPdf;
```
lennym, line 171: This should probably just care about being a PDF. There's no possible other pdf context in which we _do_ want add another buttons.

More generally it might be better to find where the `editable` prop is set and make sure it's always false for PDFs.

PR #688, opened 2021-02-03
Remove generated hint text with link from application PDF

wheelsandcogs:
i.e. remove this:

hint

### DIFF
--- a/client/components/review-field.js
+++ b/client/components/review-field.js
@@ -244,8 +244,8 @@ class ReviewField extends React.Component {
       }
       return <dl className="inline">
         {
-          species.map(s => (
-            <Fragment key={s.key}>
+          species.map((s, i) => (
+            <Fragment key={i}>
               <dt>{s.title}:</dt>
               <dd>{s.value ? s.value : <em>No answer provided.</em>}</dd>
             </Fragment>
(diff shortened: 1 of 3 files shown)


client/components/review-field.js
```diff
@@ -244,8 +244,8 @@ class ReviewField extends React.Component {
       }
       return <dl className="inline">
         {
-          species.map(s => (
-            <Fragment key={s.key}>
+          species.map((s, i) => (
+            <Fragment key={i}>
```
wheelsandcogs, line 248: unrelated fix: s.key is `undefined` here and was throwing a react warning.

client/pages/sections/objectives/review.js
```diff
@@ -19,7 +19,7 @@ const ObjectivesReview = ({ playback, values, steps, goto, readonly, isFullAppli
         </Fragment>
       )
     }
-    <Playback playback={playback} />
+    <Playback playback={playback} isFullApplicationPdf={isFullApplication && pdf} />
```
lennym, line 22: Is this potentially another place where we don't what kind of pdf it is and so should keep it simple and just pass the pdf prop?

I think yes.

lennym: Also probably _don't_ actually pass it in at all if it can be loaded from the state in the child component.

PR #694, opened 2021-02-23
Fix images not displaying in docx downloads

wheelsandcogs:
The condition `not object OR not string` caused the function to always return early and skip the code below.


lennym:
Whoops!

PR #695, opened 2021-02-23
Fix species not rendering in docx.

wheelsandcogs:


lennym:
I think because we now support object values for rich text components that there is no need to `stringify` the value at all, and we can remove the condition.

However, I'm also not sure which of the two branches where this code has been added was causing the bug. Was it being stringified when it shouldn't, or vice-versa?

client/components/download-link/renderers/docx-renderer.js
```diff
@@ -875,16 +875,19 @@ export default (application, sections, values, updateImageDimensions) => {
...
         try {
           value = JSON.parse(value);
+          valueWasJson = true;
         } catch (e) {
           return Promise.resolve(value);
         }
       }
 
       if (!value.document || !value.document.nodes) {
-        return Promise.resolve(value);
```
wheelsandcogs, line 887: If the value was a JSON string, and get's parsed to an object on line 880, then returning the object and not the original value here was causing the species not to render.

I don't know if line 901 is also causing similar issues, but I thought that it was best to always return the value in the original format.

wheelsandcogs:
> I think because we now support object values for rich text components that there is no need to `stringify` the value at all, and we can remove the condition.
> 
> However, I'm also not sure which of the two branches where this code has been added was causing the bug. Was it being stringified when it shouldn't, or vice-versa?

Most of the code from the original PR (https://github.com/UKHomeOffice/asl-projects/pull/658/files) was never actually executing because of the condition `(typeof value !== 'object' || typeof value !== 'string' || value === null)` which was always true and exited the function early, so image dimensions were never added to the word doc and the images stopped displaying.

My "fix" here (https://github.com/UKHomeOffice/asl-projects/pull/694/files) was just to correct the condition to allow the code below to actually execute, which resulted in images re-appearing, but also resulted in revealing the bug with species not rendering appearing.

The second "fix" here seems to be to always return the original value type, (i.e. not a parsed object if the original value was a JSON string).

lennym:
Yes, I'm curious to know why it was affecting other species values in particular. What format are they in?

wheelsandcogs:
Species looks like:

```
"species": [
        {
          "id": "f505671e-ccb1-496e-8efb-58c5e84afec1",
          "speciesId": "28",
          "other-species-type": "JABU"
        }
      ],
```

lennym:
I'm still not sure I see where the buggy behaviour was coming from, or how this fixes it. That value isn't a string, so `valueWasJson` stays false, and so this is a no-op since it will continue to return `Promise.resolve(value)`.

I'm not sure what I'm missing...

lennym:
The cause of this issue was the species id in particular which was being converted to a number by `JSON.parse` and needs to be a string for some downstream code that does an `id === '28'` to check for "other species".

By passing it back through `JSON.stringify` we ensure this remains the case.

PR #700, opened 2021-03-02
Show POLEs name and details when reviewing

wheelsandcogs:


wheelsandcogs:
This took a while because I didn't understand what 'enable' meant. Hoping renaming it brings a bit more clarity, but still not 100% sold on `repeaterFor`, so if you can think of a better name, please suggest.

PR #706, opened 2021-03-05
Simplify change flags for nested repeaters

lennym:
Because the changes for repeaters are calculated to include paths at each level e.g. `protocols`, `protocols.<id>`, `protocols.<id>.steps` etc then there's no need to try to enumerate the fields inside these sections - which was being done incorrectly previously - and can simply pass the path to the section level into the changed indicator component.

This vastly reduces the complexity of calculation needed to be done to establish if a repeating section has changed.


wheelsandcogs:
Screenshot 2021-03-05 at 15 10 49 like it 👍

PR #718, opened 2021-04-16
only show close section link when label set and not pdf

wheelsandcogs:


client/pages/sections/protocols/sections.js
```diff
@@ -137,7 +137,13 @@ const ProtocolSections = ({ sections, protocolState, editable, newComments, ...p
   <Accordion open={getOpenSection(protocolState, editable, sections)} toggleAll={!props.pdf} pdf={props.pdf}>
     {
       sectionNames.map((section, sectionIndex) => (
-        <ExpandingPanel key={section} title={getTitle(sections[section], newComments, props.values)} closeLabel={`Close ${sections[section].title.toLowerCase()}`} className={section.toLowerCase()}>
+        <ExpandingPanel
+          key={section}
+          title={getTitle(sections[section], newComments, props.values)}
+          className={section.toLowerCase()}
+          closeLabel={`Close ${sections[section].title.toLowerCase()}`}
```
lennym, line 144: Not sure it's as simple as slapping a `.toLowerCase()` on it because of POLEs and GAAs

PR #722, opened 2021-05-06
Fix diff window display of animal quantities

lennym:
Where a project uses "other" types these are not being correctly rendered in the diff windows for the animal quantities section. Update `mapAnimalQuantities` function to include "other" types.


lennym:
## Before:
Screenshot 2021-05-06 at 12 22 21

## After:
Screenshot 2021-05-06 at 12 23 05

PR #745, opened 2021-07-02
bugfix - RTE hints

joefitter:
* React markdown now uses children rather than source
* Updated all occurances of source


lennym:
We may want to:

a) standardise on the `<Markdown>` component from `@asl/components`
b) modify that component so that if it is passed a `source` prop it maps that to children downstream

PR #746, opened 2021-07-15
Bugfix - restore images

joefitter:
* images are blocks, they do not have text, we still want to render them


client/components/editor/blocks/render-block.js
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import { Button } from '@ukhomeoffice/react-components';
 
 function Image(props) {
+  console.log('RENDERING IMAGE')
```
wheelsandcogs, line 5:
```suggestion
```

PR #756, opened 2021-09-09
WIP react-addons-css-transition-group was moved to react-transition-group

wheelsandcogs:


wheelsandcogs:
Will revisit when node 16 becomes LTS shortly.

PR #765, opened 2021-11-01
Feature - preview

joefitter:
* Added preview component
* updated components to not be editable if in preview mode
* Added no conditions label

### DIFF
--- a/client/components/conditions/conditions.js
+++ b/client/components/conditions/conditions.js
@@ -151,25 +151,27 @@ class Conditions extends Component {
     return (
       <div className="conditions">
         {
-          // ra conditions used to be added to the conditions array, we dont want to show them here.
-          conditions.filter(c => !c.key.match(/^retrospective-assessment/)).map((condition, index) => {
-            const template = get(CONDITIONS[this.props.scope], condition.path, {});
-            const { title, content } = template;
-            return <Condition
-              key={condition.key}
-              className={condition.key}
-              singular={this.props.singular}
-              editConditions={editConditions}
-              updating={updating}
-              title={title}
-              number={index + 1}
-              custom={condition.custom}
-              content={content}
-              {...condition}
-              onSave={this.handleSave(condition.key)}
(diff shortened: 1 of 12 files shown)


client/components/preview-licence.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+
+export default function NextSteps() {
+  const { isActionable, previewLink } = useSelector(state => state.application);
+  if (!isActionable) {
```
lennym, line 6: Is there some aspect of how this is set that limits it to ASRU users only?

joefitter: It's the same logic as the next steps button, so it shows when you are looking at the version in the task (not asru specific)

joefitter: do we want it to be asru only?

lennym: The ticket is titled "Give ASRU a preview of the draft licence for PPL applications" so it's somewhat ambiguous. My assumption was that it was meant to be exclusive though.

joefitter: ok cool I'll make it asru only for now - easy to reverse

client/components/conditions/conditions.js
```diff
@@ -151,25 +151,27 @@ class Conditions extends Component {
...
-              content={content}
-              {...condition}
-              onSave={this.handleSave(condition.key)}
-              onRemove={this.handleRemove(condition.key)}
-            />
-          })
+          conditions.length === 0
+            ? <p><em>No {this.props.authorisations ? 'authorisations' : 'conditions'} added</em></p>
+            // ra conditions used to be added to the conditions array, we dont want to show them here.
+            : conditions.filter(c => !c.key.match(/^retrospective-assessment/)).map((condition, index) => {
```
lennym, line 157: Should this filter be applied before the length is checked ☝️?

joefitter: good point - updated

PR #780, opened 2022-02-08
Make horizontal table scroll more obvious for readonly tables

wheelsandcogs:
It was not obvious that some tables were wider than the space available and therefore needed scrolling.

This adds an always-on horizontal scrollbar if the table overflows it's container.

Supported: Edge, Chrome, Safari
Not supported: Firefox

Needs:
 - https://github.com/UKHomeOffice/asl-components/pull/246

### DIFF
--- a/assets/scss/projects.scss
+++ b/assets/scss/projects.scss
@@ -1153,6 +1153,7 @@ html.modal-open, html.modal-open body {
 
       table {
         color: $govuk-text-colour;
+        margin-bottom: 0;
       }
 
       .grey {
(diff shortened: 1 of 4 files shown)


assets/scss/projects.scss
```diff
@@ -1153,6 +1153,7 @@ html.modal-open, html.modal-open body {
 
       table {
         color: $govuk-text-colour;
+        margin-bottom: 0;
```
lennym, line 1156: Should this declaration be applied in the context of the overflow wrapper, so whenever the scrollbar is shown it always sits flush to the bottom of the table, or do we want to keep the gap when it's not against the purple background?

PR #789, opened 2022-05-12
Convert comments and duration components to functional component

lennym:


client/components/field.js
```diff
@@ -152,8 +152,6 @@ class Field extends Component {
         label={ label }
         hint={ hint }
         error={ this.props.error }
-        min={ this.props.min }
-        max={ this.props.max }
```
wheelsandcogs, line 156: What's the deal here?

lennym: The `Duration` component doesn't use those props. They _probably_ didn't get set anyway, but if they did they were a no-op.

PR #790, opened 2022-05-16
Removed AA should not be selectable as a protocol location

wheelsandcogs:


client/components/location-selector.js
```diff
@@ -1,29 +1,27 @@
 import React from 'react';
-import { connect } from 'react-redux';
+import { useSelector } from 'react-redux';
 import { Link } from 'react-router-dom';
-
 import Field from './field';
 import getLocations from '../helpers/get-locations';
```
wheelsandcogs, line 6: This did have some behaviour changes initially but they were reverted - decided to keep the refactor to using hooks.

client/pages/sections/granted/protocol-establishments.js
```diff
@@ -1,11 +1,13 @@
 import React from 'react';
```
wheelsandcogs, line 1: As above, this did have some behaviour changes initially but they were reverted - decided to keep the refactor to using hooks.

lennym:
I'm ok with this, but I'm unsure if it solves the underlying issue of locations which have already been selected on protocols not being removed if removed from the parent list without requiring some interaction with the protocol field itself.

i.e. if I remove an AA establishment which was previously selected on a protocol, is it automatically removed from the protocol in the _data_, and not just in the UI?

wheelsandcogs:
The AA location is removed from the protocols data and UI as soon as you remove the AA in the establishments section.

If you restore the AA, it will then be available again in the protocols location selector but it will not be ticked if it previously was.

PR #801, opened 2022-06-29
Improve messaging when adding additional establishments in PPL application

MaxOSullivan:
- Make establishmentSelector conditional based on whether it is in a new app or amendment
- Make Warning conditional based on if the element is in an application or review

This fixes the broken integration tests


client/pages/sections/establishments/establishments.js
```diff
@@ -43,7 +43,7 @@ function Establishment(props) {
           editable && <a href="#" className={classnames('inline-block float-right', { restore: values.deleted })} onClick={values.deleted ? restoreItem : removeItem}>{values.deleted ? 'Restore' : 'Remove'}</a>
         }
         <h2>{`Additional establishment ${number + 1}`}</h2>
-        <Warning className="larger">This establishment will also need to conduct an AWERB review of this application before it can be submitted.</Warning>
+        { props.actualProject && <Warning className="larger">This establishment will also need to conduct an AWERB review of this application before it can be submitted.</Warning> }
```
wheelsandcogs, line 46: `actualProject` might not be the best option here - does `editable` work?

PR #805, opened 2022-07-06
Improve messaging when adding additional establishments in PPL application

MaxOSullivan:
- Add new content based on designs below
- Added dynamically updating establishment name if the primary establishment is updated on the previous page
- If it is a draft application it uses the existing transferDraft behaviour when changing the primary establishment, if it is an amendment it goes to the previous page where it allows the user to change the primary establishment

Remade PR from https://github.com/UKHomeOffice/asl-projects/pull/801 as this one was missing content

Screenshot of added content, designs found here: https://w3iqpx.axshare.com/#id=36lndd&p=page_1

Screenshot 2022-07-06 at 17 33 40

When the user chooses to update the primary establishment this is shown on the next page

Screenshot 2022-07-06 at 17 34 15
Screenshot 2022-07-06 at 17 34 44

When they click 'Change' from this page they are returned to the previous screen as it is an amendment

### DIFF
--- a/client/pages/sections/establishments/index.js
+++ b/client/pages/sections/establishments/index.js
@@ -25,13 +25,14 @@ export default function Index({ advance, exit, ...props }) {
           <legend className="govuk-fieldset__legend">
             <h2 className="govuk-fieldset__heading govuk-heading-l larger">Primary establishment</h2>
           </legend>
-          {
-            transferToEstablishmentName ? <p className="larger">{transferToEstablishmentName}</p> : <p className="larger">{establishment.name}</p>
-          }
-          {
-            isActive ? <p className="larger"><Link to="/" label="Change"/></p>
-              : <p className="larger"><Link page="project.transferDraft" label="Change" establishmentId={establishment.id} projectId={project.id} /></p>
-          }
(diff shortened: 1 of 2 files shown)


client/pages/sections/establishments/index.js
```diff
@@ -1,11 +1,39 @@
...
     <div className="establishments-section">
       <h1>Establishments</h1>
+      <p className="larger">Add any additional establishments where work on this project will be carried out beyond just the primary establishment.</p>
+      <div className="govuk-form-group">
+        <fieldset className="govuk-fieldset inline smaller">
+          <legend className="govuk-fieldset__legend">
+            <h2 className="govuk-fieldset__heading govuk-heading-l larger">Primary establishment</h2>
+          </legend>
+          {
+            transferToEstablishmentName ? <p className="larger">{transferToEstablishmentName}</p> : <p className="larger">{establishment.name}</p>
```
lennym, line 29: Minor thing, but I'd probably remove the repeated markup and ternary operator and replace this whole block with:

```
<p className="larger">{transferToEstablishmentName || establishment.name}</p>
```

PR #807, opened 2022-07-14
Build image sources dynamically with configured root URL

lennym:
Images in PDFs loaded from S3 need to have the full domain defined, so the source needs to be dynamic depending on the defined root URL.

If a `src` attribute is defined then it's almost certainly an old `data:` url, and so should be left intact.

### DIFF
--- a/client/components/editor/blocks/render-block.js
+++ b/client/components/editor/blocks/render-block.js
@@ -1,13 +1,18 @@
 import React from 'react';
+import { useSelector } from 'react-redux';
 import { Button } from '@ukhomeoffice/react-components';
 
-function Image(props) {
+function Image({ src, token, remove, loading }) {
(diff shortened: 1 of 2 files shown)


client/components/editor/blocks/render-block.js
```diff
@@ -1,13 +1,18 @@
 import React from 'react';
+import { useSelector } from 'react-redux';
 import { Button } from '@ukhomeoffice/react-components';
 
-function Image(props) {
+function Image({ src, token, remove, loading}) {
```
wheelsandcogs, line 5:
```suggestion
function Image({ src, token, remove, loading }) {
```

wheelsandcogs: 🤔 thought this would be a linting fail.

PR #809, opened 2022-07-20
Add fallback message for users which cannot change the main establishment

MaxOSullivan:


lennym:
I hate it, but I think that's more of a me problem.

PR #818, opened 2022-08-25
Fix bug which doesnt let you edit default conditions

MaxOSullivan:
Previously the content didn't update to match the state, so after every key press the content would be set to the default content. 
Resets the content based on the state so it can be updated.

### DIFF
--- a/client/components/editable.js
+++ b/client/components/editable.js
@@ -15,6 +15,7 @@ function Editable({ edited,
   onRevert = () => {} }) {
 
   const [ state, setState ] = useState({content, reminders});
+  content = state.content;
 
   useEffect(() => {
     onChange(state);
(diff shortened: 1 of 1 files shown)


client/components/editable.js
```diff
@@ -15,6 +15,7 @@ function Editable({ edited,
   onRevert = () => {} }) {
 
   const [ state, setState ] = useState({content, reminders});
+  content = state.content;
```
wheelsandcogs, line 18: This might need to be a bit more involved - in the cancel function we're checking
```if (state.content !== content) {```
which if we do this, will never return true.

I think the intention was to prevent you navigating away from the page before the content has saved. The save is triggered by calling the parent `onChange` in the use effect, so we might want to wait for that to return before updating the content?

something like
```Promise.resolve()
  .then(() => onChange(state))
  .then(() => content = state.content);
```

PR #822, opened 2022-09-05
Add reminders with conditions 

MaxOSullivan:
Allows users to add reminders to conditions as the condition is being created

Screenshot 2022-09-05 at 15 24 09

Screenshot 2022-09-05 at 15 24 32

Screenshot 2022-09-05 at 15 24 54

### DIFF
--- a/client/components/conditions/add-conditions.js
+++ b/client/components/conditions/add-conditions.js
@@ -121,6 +121,7 @@ function AddConditions({
           onAdd={handleOnAdd}
           onUpdate={handleOnUpdate}
           onRemove={handleOnRemove}
+          // The conditionKey below is used as a temporary placeholder until a real UUID is generated when the condition is saved
           conditionKey={'uuid'}
         />
       }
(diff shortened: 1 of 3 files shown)


client/components/conditions/add-conditions.js
```diff
@@ -118,9 +121,10 @@ function AddConditions({
           onAdd={handleOnAdd}
           onUpdate={handleOnUpdate}
           onRemove={handleOnRemove}
+          conditionKey={'uuid'}
```
lennym, line 124: What does this do?

MaxOSullivan: This gives the custom condition a temporary condition key so a reminder can be saved against it in the state, then in the save function it is replaced when the UUID is actually created. Initially I tried to give it a real UUID at this point but it generated a new one every time it was rendered which broke things, so I opted for this way.

lennym: I think I see. The eventual uuid is generated at L142, but the downstream logic in the editable component wants a `conditionKey` to be set to allow reminders to be added. Is that right?

If so, could you just add a comment to the effect that the string being passed is an arbitrary placeholder and there's nothing special about the value `'uuid'`?

MaxOSullivan: Correct. Have added a comment above the line

PR #826, opened 2022-09-09
Task/update linting rules

wheelsandcogs:


.eslintrc
```diff
@@ -1,10 +1,8 @@
 extends:
-  - eslint:recommended
-  - plugin:react/recommended
+  - lennym
```
lennym, line 2: All your base are belong to this guy.

PR #829, opened 2022-09-21
Removing species from project level should flag protocol as changed

wheelsandcogs:
When saving changes to the project level species, this will now strip the species from `protocol.species` list, but leave the `protocol.speciesDetails` intact.

This will render a changed badge immediately after save against the Protocols section and the relevant protocols.

### DIFF
--- a/client/pages/sections/protocols/animals.js
+++ b/client/pages/sections/protocols/animals.js
@@ -18,18 +18,6 @@ import Repeater from '../../../components/repeater';
 
 const allSpecies = flatten(Object.values(SPECIES));
 
-function getProjectSpecies(project) {
-  return flatten([
-    ...(project.species || []).map(s => {
-      if (s.indexOf('other') > -1) {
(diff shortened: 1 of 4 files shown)


client/pages/sections/protocols/animals.js
```diff
@@ -46,10 +34,9 @@ function normaliseValues(speciesDetails) {
 
 export function filterSpeciesByActive(protocol, project) {
   const { species = [], speciesDetails = [] } = protocol;
-  const projectSpecies = getProjectSpecies(project);
 
   return normaliseValues(speciesDetails).filter(s => {
-    return (species.includes(s.value) || species.includes(s.name)) && (projectSpecies.includes(s.value) || projectSpecies.includes(s.name));
```
lennym, line 52: We may need to keep this for backwards compatibility with old data where the protocols still have species included in data that are not on the top level project.

They won't be removed until they're edited.

PR #831, opened 2022-09-26
Add new cosmetic condition and update existing conditions

MaxOSullivan:


lennym:
Because this needs some communications from ASRU I'm going to leave it here until at least Thursday. It's fine though.

lennym (approved):

PR #833, opened 2022-10-03
ASL-4046 Reusable steps

gaz1634:
This will now store reusable steps in a property at the top level of the project-versions:
```
"reusableSteps": {
  "39b40f1d-0cb3-47b2-8ebe-0306ec2a5ba0": {
    "id": "39b40f1d-0cb3-47b2-8ebe-0306ec2a5ba0",
    "saved": true,
    "title": {
    ...
    },
    "adverse": true,
    "optional": false,
    "reusable": true,
    "completed": true,
  }
}
```

Where a step within the protocols is using a re-usable step, it will only include the id, and a reusableStepId:
```
"steps": [
  {
  "id": "39b40f1d-0cb3-47b2-8ebe-0306ec2a5ba0",
  "reusableStepId": "39b40f1d-0cb3-47b2-8ebe-0306ec2a5ba0"
  }
]

```
**Editing step:**
Now shows new step reference and re-use question

Steps editing - Edit

___

**Adding a new step**
When there is at least one re-usable step, then when adding a step the following option is available:

Add new step selector

___

**View Step:** 
This now has an option to:
Edit the step, which creates a new step with details completed and the original details in existingValues field so the chance can be cancelled.
Edit every instance, which creates a copy of the values in existingValues for cancelling.

Steps editing - View

___

**Review screen**
Changed to use expandable components, with the reference, or first line of text if no reference provided

Review Expandable

___

**Granted Licence**

Now displays the reference in the index and step header if it is present.
Granted Licence

### DIFF
--- a/client/actions/projects.js
+++ b/client/actions/projects.js
@@ -11,6 +11,7 @@ import sendMessage from './messaging';
 import { getConditions } from '../helpers';
 import cleanProtocols from '../helpers/clean-protocols';
 import sha from 'sha.js';
+import { keyBy } from 'lodash';
 
 const CONDITIONS_FIELDS = ['conditions', 'retrospectiveAssessment'];
 
@@ -87,6 +88,17 @@ export function updateProject(project) {
   return {
     type: types.UPDATE_PROJECT,
     project
+  };
+}
+
+export function saveReusableSteps(reusableSteps) {
+  return (dispatch, getState) => {
+    const state = getState();
+    const newState = cloneDeep(state.project);
+    const updatedReusableSteps = keyBy(reusableSteps, 'id');
+    newState.reusableSteps = {...newState.reusableSteps, ...updatedReusableSteps};
+    dispatch(updateProject(newState));
+    return debouncedSyncConditions(dispatch, getState);
   };
 }
 
--- a/client/actions/projects.js
+++ b/client/actions/projects.js
@@ -98,7 +98,7 @@ export function saveReusableSteps(reusableSteps) {
     const updatedReusableSteps = keyBy(reusableSteps, 'id');
     newState.reusableSteps = {...newState.reusableSteps, ...updatedReusableSteps};
     dispatch(updateProject(newState));
-    return debouncedSyncConditions(dispatch, getState);
+    return debouncedSyncProject(dispatch, getState);
   };
 }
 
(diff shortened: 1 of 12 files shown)


client/actions/projects.js
```diff
@@ -90,6 +91,17 @@ export function updateProject(project) {
...
 }
 
+export function saveReusableSteps(reusableSteps) {
+  return (dispatch, getState) => {
+    const state = getState();
+    const newState = cloneDeep(state.project);
+    const updatedReusableSteps = keyBy(reusableSteps, 'id');
+    newState.reusableSteps = {...newState.reusableSteps, ...updatedReusableSteps};
+    dispatch(updateProject(newState));
+    return debouncedSyncConditions(dispatch, getState);
```
lennym, line 101: Using `debouncedSyncConditions` and not `debouncedSyncProject` here feels _very_ odd to me. Is there a reason to use the conditions sync and not the normal project sync?

I'm surprised this even works given the conditions sync should make requests to an endpoint that is only accessible to ASRU users.

gaz1634: Good spot, I will fix this locally and check everything is working as expected.

It looks as though the conditions sync is only syncing if there is data there so wouldn't have called the endpoint.

lennym: I suspect it's not doing _anything_ then, and it's being correctly synced after the fact by another call to the sync function.

lennym:
One slight oddity when you have a step repeated, then repeat it across multiple protocols.

If you expand the steps on each protocol, then choose to edit the repeated step in all protocols, then the step becomes editable in all instances, not just the one you clicked to edit in. This means if you make a change, then scroll down the page you get another edit form for the step, but with the original values still in it.

client/actions/projects.js
gaz1634: This has been updated to use the `debouncedSyncProject`

client/pages/sections/protocols/steps.js
```diff
@@ -81,21 +126,101 @@ class Step extends Component {
...
+              readonly={true}
+              className={'reusable'}
+            />
+            <Warning>You cannot change this answer when editing all instances of this step.</Warning>
+          </Fragment>
+          }
+          <p className="control-panel">
+            <Button onClick={this.saveStep}>Save step</Button>
+            {
+              length > 1 && <Button className="link" onClick={this.removeItem}>Remove step</Button>
```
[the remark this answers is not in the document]
gaz1634: Updated to show the `Remove this step from this protocol` when its a reusable step

PR #836, opened 2022-10-06
Set rows on condition textareas to fit default text

MaxOSullivan:
https://collaboration.homeoffice.gov.uk/jira/browse/ASL-4107

Textareas now fit the default text when you begin editing

### DIFF
--- a/client/components/editable.js
+++ b/client/components/editable.js
@@ -66,6 +66,8 @@ function Editable({ edited,
         value={state.content}
         onChange={onContentChange}
         autoExpand={true}
+        // set the rows to fit the default text, 72 has no inherent value it just works as a good average character length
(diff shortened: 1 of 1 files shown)


client/components/editable.js
```diff
@@ -66,6 +66,8 @@ function Editable({ edited,
         value={state.content}
         onChange={onContentChange}
         autoExpand={true}
```
lennym, line 68: This _should_ be sufficient, so if it's not then we probably need to go a little deeper.

MaxOSullivan: Do you mean the autoexpand? That enables when the user makes a change but it doesn't default to the correct size

lennym: I do mean the autoexpand. It should be fixed there to make it default to the correct size.

PR #837, opened 2022-10-11
Add autofocus to condition text areas so they are automatically the correct size

MaxOSullivan:
Add autoFocus to Editable component
Custom conditions contain no content and are loaded before they are enabled by the user so added a conditional to not enable auto focus

### DIFF
--- a/client/components/editable.js
+++ b/client/components/editable.js
@@ -66,6 +66,8 @@ function Editable({ edited,
         value={state.content}
         onChange={onContentChange}
         autoExpand={true}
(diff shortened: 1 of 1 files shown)


client/components/editable.js
```diff
@@ -66,6 +66,8 @@ function Editable({ edited,
         value={state.content}
         onChange={onContentChange}
         autoExpand={true}
+        // do not add autofocus to the custom condition as it is loaded before you enable it
+        autoFocus={state.content !== ''}
```
lennym, line 70: Is it possible for this to result in there being multiple autofocused textareas on the same page at the same time?

PR #839, opened 2022-10-13
Only edit one reusable step instance at a time

lennym:
Adds a bit of logic that if it's a reusable step being edited then only the one instance that was clicked on shows as editable and the other instances stay read-only.

This code could use a serious refactor to stop confusing `editable` and `completed` tbh, but not now.

### DIFF
--- a/client/pages/sections/protocols/steps.js
+++ b/client/pages/sections/protocols/steps.js
@@ -92,7 +92,7 @@ class Step extends Component {
 
   cancelItem = e => {
     e.preventDefault();
-    this.props.updateItem({ ...(cloneDeep(this.props.values.existingValues)), existingValues: undefined });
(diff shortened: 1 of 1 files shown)


client/pages/sections/protocols/steps.js
```diff
@@ -92,7 +92,7 @@ class Step extends Component {
 
   cancelItem = e => {
     e.preventDefault();
-    this.props.updateItem({ ...(cloneDeep(this.props.values.existingValues)), existingValues: undefined });
+    this.props.updateItem({ ...(cloneDeep(this.props.values.existingValues)), existingValues: undefined, completed: true });
```
lennym, line 95: Otherwise things get a bit weird if you edit and cancel

PR #841, opened 2022-10-17
ASL-4046 - Ensure protocol.steps is an array - for legacy projects

gaz1634:


client/helpers/steps.js
```diff
@@ -62,7 +62,11 @@ export const getStepTitle = (title, untitled = <em>Untitled step</em>) => {
     : untitled;
 };
 
-export const reusableStepFieldKeys = (protocol) =>
-  (protocol.steps || [])
+export const reusableStepFieldKeys = (protocol) => {
+  if (!Array.isArray(protocol.steps)) {
+    return [];
+  }
+  return (protocol.steps || [])
```
lennym, line 69: What values did it have before that weren't caught by `protocol.steps || []`?

gaz1634: On the legacy project used in the autoproject it, protocol.steps seems to be an object, so the filter function does not exist.

lennym: That's fine. If it's a seed data problem rather than an actual "data in the wild has ended up here" problem then I'm more relaxed about it.

Just if real prod data had got itself into an unexpected state we should find out why.

PR #845, opened 2022-10-19
ASL-4046 - Include reference + repeated from for re-usable/re-used step headers on 'View latest submission' playback

gaz1634:
I missed the View latest submission playback when adding the reference and 'repeated from protocol X' to the step headers.

Screenshot 2022-10-19 at 08 27 59

### DIFF
--- a/client/pages/sections/protocols/steps.js
+++ b/client/pages/sections/protocols/steps.js
@@ -139,7 +139,8 @@ class Step extends Component {
       protocol,
       newComments,
       reusableSteps,
-      pdf
+      pdf,
+      readonly
     } = this.props;
     const changeFieldPrefix = values.reusableStepId ? `reusableSteps.${values.reusableStepId}.` : this.props.prefix;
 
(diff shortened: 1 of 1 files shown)


client/pages/sections/protocols/steps.js
```diff
@@ -242,14 +243,17 @@ class Step extends Component {
...
               completed && !isUndefined(values.optional) &&
               <span className="light smaller">{` (${values.optional === true ? 'optional' : 'mandatory'})`}</span>
             }
+            {
+              !pdf && readonly && repeatedFrom && <div className="light smaller">{`Repeated from protocol ${repeatedFrom}`}</div>
+            }
           </h3>
           {
-            pdf && repeatedFrom && <span className="review"><p className="grey">{`Repeated from protocol ${repeatedFrom}`}</p></span>
+            pdf && !readonly && repeatedFrom && <span className="review"><p className="grey">{`Repeated from protocol ${repeatedFrom}`}</p></span>
```
lennym, line 256: I'm not sure where the props come from, but `pdf && !readonly` implies to me that we're in an editable PDF state, which I find confusing.

gaz1634: Removed, the `!readonly` wasn't necessary when it is a pdf, and as you say it is confusing.

PR #847, opened 2022-10-21
Default checksum omission to `id`

lennym:
The version data is extended on the client with the id, which means it will always differ from the server value, which does not have an id value.

This prevents an error being thrown in RAs where no omissions are defined on the server side because an RA does not include any server computed properties.

Also improve error handling so that any client side errors in the sync pipeline are correctly logged. Currently if there is an error thrown during the sync pipeline (such as the above) it is not reported.

### DIFF
--- a/client/actions/projects.js
+++ b/client/actions/projects.js
@@ -8,7 +8,7 @@ import shasum from 'shasum';
 import * as types from './types';
 import database from '../database';
 import { showMessage, showWarning, throwError } from './messages';
-import sendMessage from './messaging';
+import sendMessage, { postError } from './messaging';
 import { getConditions } from '../helpers';
 import cleanProtocols from '../helpers/clean-protocols';
 import sha from 'sha.js';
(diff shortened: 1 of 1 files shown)


client/actions/projects.js
```diff
@@ -250,16 +250,18 @@ const shouldSyncProject = state => {
...
   dispatch(syncError());
   const errorCount = getState().application.errorCount;
-  if (err.code === 'UPDATE_REQUIRED') {
+  if (error.code === 'UPDATE_REQUIRED') {
     return dispatch(throwError('This software has been updated. You must refresh your browser to avoid losing work.'));
   }
   if (errorCount > 5) {
     return dispatch(throwError('Failed to save your changes. Try refreshing your browser to continue. If the problem persists then please report to aspeltechnicalqueries@homeoffice.gov.uk'));
   }
+  console.error(error);
```
wheelsandcogs, line 263: intentional?

lennym: Intentional.

client/actions/projects.js
```diff
@@ -321,7 +323,10 @@ const syncProject = (dispatch, getState) => {
     })
     .then(response => {
       const patched = jsondiff.patch(state.savedProject, patch);
-      if (response.checksum !== shasum(omit(patched, ...response.checksumOmit))) {
+      // always exclude the id from the checksum since it does not exist in the server data and is set locally
+      // also exclude any server-computed properties that are returned in the response
+      const checksumOmit = [...(response.checksumOmit || []), 'id'];
+      if (response.checksum !== shasum(omit(patched, ...checksumOmit))) {
```
wheelsandcogs, line 329: `omit()` can take an array as the 2nd param so no need for the expansion here really.

PR #853, opened 2022-10-27
Disable hidden checkbox to stop translation research being displayed twice

MaxOSullivan:
The bug https://collaboration.homeoffice.gov.uk/jira/browse/ASL-4119 is caused by users clicking a hidden checkbox which then causes flattenReveals to run on the translational research values, giving it it's own section when being reviewed.

Disabling the hidden checkbox stops it from being added to values and will stop this bug happening in the future.
Added styling to stop the disabled checkbox losing opacity

### DIFF
--- a/assets/scss/projects.scss
+++ b/assets/scss/projects.scss
@@ -2,6 +2,7 @@
 @import "images";
 $light-green: #26A197;
 $pink: #ff1493;
+$hint-colour: #696F73;
 
 @mixin sticky {
   position: -webkit-sticky;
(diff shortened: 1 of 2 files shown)


assets/scss/projects.scss
```diff
@@ -2,6 +2,7 @@
 @import "images";
 $light-green: #26A197;
 $pink: #ff1493;
+$hint-colour: #696F73;
```
gaz1634, line 5: Is this change needed?

MaxOSullivan: The hint colour is used at the bottom of the css file but never defined so I was getting an error when I built it in projects, so this fixes that. Sorry should have also mentioned this in the PR

PR #859, opened 2022-11-07
Bugfix/scrollable tables

MaxOSullivan:
Add scrollable element to tables in rich text editors
Set cell widths to make the tables look better
Add CSS to new div to ensure sizing of the editor is unchanged

The table works on its own axis and scrolling doesn't effect the information around it.

Before
Screenshot 2022-11-07 at 12 18 26

After
Screenshot 2022-11-07 at 12 21 45

This also adds a scroller to the review page.
Screenshot 2022-11-07 at 13 48 01


lennym:
👏

PR #894, opened 2023-11-06
Revert "Bugfix/assb 1345 without diff window regression"

taomoh:
Reverts UKHomeOffice/asl-projects#893

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@asl/projects",
-  "version": "14.1.5",
+  "version": "14.1.3",
   "description": "ASL PPL prototype",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@asl/projects",
-  "version": "14.1.4",
+  "version": "14.1.3",
```
jeff-horton-ho-sas, line 3: Will this work with newer versions already published to npm - should we instead be releasing a newer version with the reverted code?

taomoh: Changed package version as per discussion

jeff-horton-ho-sas (approved):

PR #910, opened 2024-06-25
conditional render and hyperlink done.

farrukhishaq:


mohomeoffice (commented):
there seems to lots of change is it worth breaking it down to small changes ?

PR #911, opened 2024-07-03
feat: [ASL-4532] Add redux dev tools

kiran-varma-home-office:


jeff-horton-ho-sas (approved):

client/store.js
```diff
@@ -1,13 +1,23 @@
...
 import thunk from 'redux-thunk';
+import { composeWithDevTools } from '@redux-devtools/extension';
 
 const middlewares = [thunk];
 
 export default function configureStore(initialState = {}) {
+
+  let middleware;
+
+  if (process.env.NODE_ENV === 'development') {
```
farrukhishaq, line 12: new entry to `.env.example`
NODE_ENV === 'development'

farrukhishaq (commented):
Looks good.

client/store.js
kiran-varma-home-office: that's in different repo right? - asl-internal-ui

I guess I raise another PR for it?

jeff-horton-ho-sas: There will need to be a cascade of PRs - so after this there will need to be a commit to asl-pages bumping the asl-projects version, then to asl and asl-internal-ui bumping the pages and projects versions. You can update the .example.env at the same time.

PR #916, opened 2024-07-15
Adding code to highlight changes done for radio selection.

parvinba:
https://collaboration.homeoffice.gov.uk/jira/browse/ASL-4533

### DIFF
--- a/client/components/diff-window.js
+++ b/client/components/diff-window.js
@@ -163,6 +163,20 @@ const DiffWindow = (props) => {
         });
     };
 
+    const radioDiff = () => {
+      return (
+          <p>  
+            {
+              value === undefined ? (
+                  <em>{ DEFAULT_LABEL }</em>
+              ) : ( 
+                <span className={`diff ${parts.added ? 'added': 'removed'}`}> { value ?  'Yes' : 'No'}</span> 
+              )
+            }
+          </p>
+      );
+    };
+
     const permissiblePurposeDiff = () => {
       const diffs = parts
         .reduce((arr, { value, added, removed }) => {
@@ -221,6 +235,8 @@ const DiffWindow = (props) => {
               <em>{ DEFAULT_LABEL }</em>
             </p>
           );
+      case 'radio':
+        return radioDiff();
       case 'permissible-purpose':
         return parts.length
           ? (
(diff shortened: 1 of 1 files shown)


client/components/diff-window.js
```diff
@@ -304,6 +320,12 @@ const DiffWindow = (props) => {
   const compare = () => {
 
     const hasVisibleChanges = hasContentChanges(before, props.value, props.type);
+    let removed = changes.removed;
+    let added = changes.added;
```
kiran-varma-home-office, line 324:
```suggestion
    let { removed, added } = changes;
```

client/components/diff-window.js
```diff
@@ -304,6 +320,12 @@ const DiffWindow = (props) => {
...
+    if (props.type === 'radio') {
+      removed = { added: false };
+      added = { added: true };
+    }
```
kiran-varma-home-office, line 328: compare function has only render at the moment, I am thinking maybe its good to move above to renderDiff !

also, I don't see `removed = { added: false };` used anywhere, am I missing something !?

parvinba: It is too late for renderdiff to determine if it is added or removed. Existing comparison for checkbox does it here. Removed is passed in previous window (line 347)

client/components/diff-window.js
parvinba: done..

client/components/diff-window.js
kiran-varma-home-office: just a thought, i think you could remove all this and do everything in radioDiff function you created .. would this work? ( I didn't test it, I just typed quickly)

`        const radioDiff = (parts) => {
      const newParts = { ...parts, removed: { added: false }, added: { added: true }}
      return (
        <p>
          {
            value === undefined ? (
              <em>{ DEFAULT_LABEL }</em>
            ) : (
              <span className={`diff ${newParts.added ? 'added' : 'removed'}`}> { value ? 'Yes' : 'No'}</span>
            )
          }
        </p>
      );
    };`

parvinba: No it will not work..

kiran-varma-home-office (approved):
I am happy to approve but please get other opinions 👍

jeff-horton-ho-sas (approved):

PR #920, opened 2024-07-16
bug: ASL-4532 Highlighting fix

kiran-varma-home-office:


client/components/diff-window.js
```diff
@@ -320,7 +320,7 @@ const DiffWindow = (props) => {
   const compare = () => {
 
     const hasVisibleChanges = hasContentChanges(before, props.value, props.type);
-    let { removed, added } = changes.removed;
+    let { removed, added } = changes;
```
kiran-varma-home-office, line 323: I tested my changes without your changes and everything seemed to work fine after fix in line 51. 

When I tested with your changes line 323 is causing issues in couple of places, above change seems to have fixed the scenarios I was testing. I don't think I have tested the scenarios related to your bug. 

Could you please check? - @parvinba

parvinba (approved):

jeff-horton-ho-sas (approved):

client/components/diff-window.js
parvinba: looks ok to me...

parvinba (approved):

PR #927, opened 2024-07-31
Asl/4495 keeping animal alive

farrukhishaq:
the protocol had a refactored previously. Protocols are rendered as VIEWABLE and EDITABLE. 
Introduced a check of editable so the `NTSFateOfAnimalFields` are rendered appropriately for each view.


jeff-horton-ho-sas (commented):
Added a comment with some refactorings I think will help, but looks good

parvinba (approved):

client/pages/sections/protocols/protocols.js
```diff
@@ -79,17 +80,18 @@ class Protocol extends PureComponent {
...
+      conditionalFateOfAnimalFields = renderFieldsInProtocol(this.props.project['fate-of-animals']);
+      // Ensure options array exists, and is initialised properly
+      if (!this.props.sections.fate.fields[0].options) {
+        this.props.sections.fate.fields[0].options = [];
+      }
+      // Update the options array with unique fields
+      this.props.sections.fate.fields[0].options = conditionalFateOfAnimalFields;
+    } else {
+      this.props.sections.fate.fields[0].options = NTSFateOfAnimalFields();
     }
```
jeff-horton-ho-sas, line 94: Some possible refactorings:
* Given we're immediatly returning based on editable or not, I think pulling the return into the if block is clearer
* `// Ensure options array exists, and is initialised properly` block is redundant as it's immediately replaced. Consider using https://docs-lodash.com/v4/set/ if you're worried the structure of the sections may not match expectations.
* The comments should be focussed on explaining why, what is happening should come from the code.
* The intent of overriding the sections passed to the component I think is clearer if we pass it in as a prop, but that might just be me.

```suggestion
    const sections = this.props.sections;
    if (editable && this.state.active) {
      // The first field in the fate section is the checkbox group to select which
      // fates apply to animals in the protocol. These are conditional on those is 
      // selected in the licences Non-technical summary.
      const conditionalFateOfAnimalFields = renderFieldsInProtocol(this.props.project['fate-of-animals']);
      this.props.sections.fate.fields[0].options = conditionalFateOfAnimalFields;
      
      return <Form
        {...this.props}
        sections={sections}
        toggleActive={this.toggleActive}
      />
    } else {
      sections.fate.fields[0].options = NTSFateOfAnimalFields();
      
      return <ProtocolSections
        {...this.props}
        sections={sections}
        newComments={newComments}
        protocolState={isActive && protocolState}
        onToggleActive={this.toggleActive}
      />;
    }
  }
```

jeff-horton-ho-sas (approved):

PR #928, opened 2024-08-02
Asl/4495 keeping animal alive

farrukhishaq:
The protocol Review page is server-side rendering. The changes will enable the content to be displayed correctly.


client/helpers/index.js
```diff
@@ -248,8 +248,16 @@ export const isTrainingLicence = values => {
...
+  }
+  return null;
+};
+
+export const UrlMarkdown = (anchoreName) => {
+  const url = getCurrentURLForFateOfAnimals();
+  if (url == null) {
+    return anchoreName;
+  }
+  return `[${anchoreName}](${url})`;
```
jeff-horton-ho-sas, line 262: Anchore should be anchor (anchore is a security tool - so this was confusing at first glance). 

possibly "linkText" would be even clearer?

```suggestion
export const UrlMarkdown = (linkText) => {
  const url = getCurrentURLForFateOfAnimals();
  if (url == null) {
    return linkText;
  }
  return `[${linkText}](${url})`;
```

jeff-horton-ho-sas (approved):
Minor typo - please update before merging, but otherwise this looks good :)

parvinba (approved):

PR #936, opened 2024-08-15
[ASL-4495] Change deselection behavior

jeff-horton-ho-sas:
- Kept-alive answers count as existing data for kept-alive, set-free, and rehomed
- Deselecting the last of kept-alive, set-free, or rehomed should be what triggers clearing kept-alive answers
- Clearing an NTS checkbox should deselect that option in all protocols (this was what was causing some to be pre-ticked in UAT demo)
- Individual protocols and the protocol section should only be marked as incomplete if there is a change.


parvinba (approved):

farrukhishaq (approved):
nice refactoring is done.

PR #941, opened 2024-08-28
Display Removed flag in protocols' steps

parvinba:


client/components/repeater.js
```diff
@@ -88,7 +88,7 @@ class Repeater extends Component {
     return Promise.resolve()
       .then(this.props.onBeforeRemove)
       .then(() => {
-        if (this.props.softDelete) {
+        if (this.props.softDelete && !this.state.items[index].reusable) {
```
jeff-horton-ho-sas, line 91: I don't think this is correct - what is the intent here? 

Note that whilst the input name is `reusable` this is transformed before the step is added to the protocol. To determine if an already created step is reusable the test would be `this.state.items[index].reusableStepId != null`

client/pages/sections/protocols/steps.js
```diff
@@ -313,14 +324,19 @@ class Step extends Component {
...
+            !values.deleted && <ChangedBadge fields={changeFields(values, changeFieldPrefix)} protocolId={protocol.id} />
+          }
           <Expandable expanded={expanded} onHeaderClick={() => onToggleExpanded(index)}>
             <Fragment>
               <p className={'toggles float-right'}>
                 <Button className="link no-wrap" onClick={() => onToggleExpanded(index)}>{expanded ? 'Close' : 'Open'} step</Button>
               </p>
               {values.reference ? <h3 className={'title inline'}>{values.reference}</h3> : <h3 className={'title no-wrap'}>{getStepTitle(values.title)}</h3>}
-              <h4 className="light">Step {index + 1} {values.optional === true ? '(optional)' : '(mandatory)'}{repeatedFrom ? ` - repeated from protocol ${repeatedFrom}` : ''}</h4>
+              <h4 className="light">Step { !values.deleted && number + 1 } {values.optional === true ? '(optional)' : '(mandatory)'}{repeatedFrom ? ` - repeated from protocol ${repeatedFrom}` : ''}</h4>
```
jeff-horton-ho-sas, line 339:
```suggestion
              <h4 className="light">{ values.deleted ? 'Removed step' : `Step ${number + 1}` } {values.optional === true ? '(optional)' : '(mandatory)'}{repeatedFrom ? ` - repeated from protocol ${repeatedFrom}` : ''}</h4>
```

parvinba: Good suggesting done!!

client/components/repeater.js
parvinba: suggested change will not work, as it does not differentiate between an instance of reusable versus reusable object itself. I added following comment in code so that intension of change is clear.
//this is to ensure that reusable step itself is not deleted, as reusable steps may have been used in other places.

parvinba: This is needed, when I create reusable step and then reused in another steps. If I delete reused steps, it also delete the initial step from protocol (so 2 steps are deleted). With this additional check it works as expected.

jeff-horton-ho-sas (approved):

PR #944, opened 2024-09-17
Steps - tagging new up and down

parvinba:


mohomeoffice (approved):
you could use  useMemo for stepIds and flattening the loops, especially with larger datasets it would enhance performance.

PR #949, opened 2024-09-24
tackling failing test cases - dev deployment

parvinba:


farrukhishaq (approved):

client/pages/sections/protocols/steps.js
```diff
@@ -247,7 +247,7 @@ class Step extends Component {
       >
         <NewComments comments={relevantComments} />
         {
-          !values.deleted && <StepBadge fields={values} changeFieldPrefix={changeFieldPrefix} protocolId={protocol.id} position={index}/>
+          !values.deleted && !pdf && <StepBadge fields={values} changeFieldPrefix={changeFieldPrefix} protocolId={protocol.id} position={index}/>
```
jeff-horton-ho-sas, line 250: The ChangeBadge code handles this by checking if the arrays are valid https://github.com/UKHomeOffice/asl-projects/blob/master/client/components/changed-badge.js#L23-L32. I don't think we can rely on all of them being set if we're not rendering a pdf - so I think it would be better to check in the StepBadge and return null if the previous step arrays aren't defined.

jeff-horton-ho-sas (changes requested):
Whilst the tests are only failing on the PDF I don't trust that to be the only edge case where this won't work. See comment.

client/components/step-badge.js
```diff
@@ -7,6 +7,9 @@ const changeFields = (step, prefix) => step.reusable ? [ `reusableSteps.${step.r
 
 export default function StepBadge(props) {
   const { previous, steps, firstSteps, grantedSteps } = useSelector(state => state.application.previousProtocols);
+  if (!previous || !steps || !firstSteps || !grantedSteps) {
+    return <></>;
```
jeff-horton-ho-sas, line 11:
```suggestion
    return null;
```

jeff-horton-ho-sas (approved):

PR #953, opened 2024-10-10
ASL-4516 AC-4b  step movement

parvinba:


mohomeoffice (changes requested):
Could you please try this - (!grantedSteps.length && firstIndex >= 0) instead of grantedSteps.length === 0 && firstIndex !== -1

mohomeoffice (approved):
looks nice to me

PR #956, opened 2024-10-17
[ASSB-1425] Kept alive should only show in protocols when kept alive is ticked in NTS

jeff-horton-ho-sas:
https://collaboration.homeoffice.gov.uk/jira/browse/ASSB-1425


mohomeoffice (approved):
looks okay to me

PR #976, opened 2025-02-19
Fix undefined error in hasDatabaseChange function

mohomeoffice:
Fix undefined error in `hasDatabaseChange` function

### DIFF
--- a/client/helpers/field-change-detection.js
+++ b/client/helpers/field-change-detection.js
@@ -13,11 +13,12 @@ import { normaliseDuration } from './normalise-duration';
  */
 export function hasDatabaseChange(fieldName, storedValue, currentValue, values, hasSpeciesFieldChanges) {
   // Prevents crashes if fieldName is undefined or invalid
-  if (!fieldName || typeof fieldName !== 'string') {
-    return false;
+  if (fieldName == null) {
+    throw new Error('fieldName must be provided.');
   }
+
+  const actualCurrentValue = currentValue ?? values?.[fieldName] ?? '';
   const adjustedStoredValue = fieldName === 'duration' ? normaliseDuration(storedValue) : storedValue;
-  const actualCurrentValue = currentValue ?? values[fieldName] ?? '';
 
   const normalisedStoredValue = fieldName === 'duration'
     ? normaliseDuration(adjustedStoredValue)
--- a/client/helpers/field-change-detection.js
+++ b/client/helpers/field-change-detection.js
@@ -12,6 +12,10 @@ import { normaliseDuration } from './normalise-duration';
  * @returns {boolean} - `true` if a change is detected, otherwise `false`.
  */
 export function hasDatabaseChange(fieldName, storedValue, currentValue, values, hasSpeciesFieldChanges) {
+  // Prevents crashes if fieldName is undefined or invalid
+  if (!fieldName || typeof fieldName !== 'string') {
+    return false;
+  }
   const adjustedStoredValue = fieldName === 'duration' ? normaliseDuration(storedValue) : storedValue;
   const actualCurrentValue = currentValue ?? values[fieldName] ?? '';
 


client/helpers/field-change-detection.js
```diff
@@ -12,6 +12,10 @@ import { normaliseDuration } from './normalise-duration';
  * @returns {boolean} - `true` if a change is detected, otherwise `false`.
  */
 export function hasDatabaseChange(fieldName, storedValue, currentValue, values, hasSpeciesFieldChanges) {
+  // Prevents crashes if fieldName is undefined or invalid
+  if (!fieldName || typeof fieldName !== 'string') {
+    return false;
+  }
   const adjustedStoredValue = fieldName === 'duration' ? normaliseDuration(storedValue) : storedValue;
   const actualCurrentValue = currentValue ?? values[fieldName] ?? '';
```
jeff-horton-ho-sas, line 20: The [error logged in opensearch](https://opensearch-notprod.acp.homeoffice.gov.uk/app/discover#/view/47c4a640-7b1e-11ef-9acc-f5490bcecafc?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15m,to:now))&_a=(columns:!(message_json.message,message_json.level,kubernetes.labels.name,kubernetes.container_name),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'1d5f13b0-b435-11ed-8e3f-810010040a12',key:kubernetes.container_name,negate:!t,params:(query:nginx-proxy),type:phrase),query:(match_phrase:(kubernetes.container_name:nginx-proxy))),('$state':(store:appState),exists:(field:message_json.level),meta:(alias:!n,disabled:!f,index:'1d5f13b0-b435-11ed-8e3f-810010040a12',key:message_json.level,negate:!f,type:exists,value:exists)),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'1d5f13b0-b435-11ed-8e3f-810010040a12',key:message_json.level,negate:!t,params:(query:info),type:phrase),query:(match_phrase:(message_json.level:info)))),index:'1d5f13b0-b435-11ed-8e3f-810010040a12',interval:auto,query:(language:kuery,query:''),sort:!())) looks to be caused by `values` being undefined/null, not `fieldName`.

```suggestion
  const actualCurrentValue = currentValue ?? values?.[fieldName] ?? '';
```

client/helpers/field-change-detection.js
```diff
@@ -12,6 +12,10 @@ import { normaliseDuration } from './normalise-duration';
  * @returns {boolean} - `true` if a change is detected, otherwise `false`.
  */
 export function hasDatabaseChange(fieldName, storedValue, currentValue, values, hasSpeciesFieldChanges) {
+  // Prevents crashes if fieldName is undefined or invalid
+  if (!fieldName || typeof fieldName !== 'string') {
+    return false;
+  }
```
jeff-horton-ho-sas, line 18: For arrays this could be a number, we can probably rely on javascript converting this to a scalar automatically rather than type checking, also it would be a code error if the fieldName was missing as that comes from the schema.
```suggestion
  if fieldName == null {
    throw new Error('fieldName must be provided.');
  }
```

jeff-horton-ho-sas (changes requested):

client/helpers/field-change-detection.js
```diff
@@ -12,8 +12,13 @@ import { normaliseDuration } from './normalise-duration';
  * @returns {boolean} - `true` if a change is detected, otherwise `false`.
  */
 export function hasDatabaseChange(fieldName, storedValue, currentValue, values, hasSpeciesFieldChanges) {
+  // Prevents crashes if fieldName is undefined or invalid
+  if (fieldName == null) {
+    throw new Error('fieldName must be provided.');
+  }
+
+  const actualCurrentValue = currentValue ?? values?.[fieldName] ?? '';
   const adjustedStoredValue = fieldName === 'duration' ? normaliseDuration(storedValue) : storedValue;
```
jeff-horton-ho-sas, line 21: This is redundant - we're calling normaliseDuration again on line 24
```suggestion
```

PR #984, opened 2025-03-11
Fix asl-service for redirect_uri parameter and SPI options depreciation in LTS keycloak.

farrukhishaq:
asl-service updated, redirect to log in on logout.


mohomeoffice (approved):
looks okay to me

